PR #16, opened 2021-02-08
add the legacy directory for the old fluid version

wawltor:
## PR Types
Files Add

## PR Changes
Add the new directory

## Describe
Adding the legacy directory that adding the fluid version in the PaddleNLP，maybe we will delete it later.


ZeyuChen (approved):
LGTM

PR #21, opened 2021-02-19
add the do_eval for the lac model train.

wawltor:
## PR Types
Function improvement

## PR Changes
Training scripts

## Describe
1.  Add the eval flag for the lac training,  the training process will not to be evaluating if set do_eval to False


ZeyuChen (approved):
lgtm

PR #69, opened 2021-03-04
Fix all incorrect segment padding

smallv0221:
### PR types

Bug fixes
### PR changes

APIs 
### Describe

Fix all incorrect segment padding


ZeyuChen (approved):
LGTM

PR #101, opened 2021-03-10
add start training from checkpoint in bigbird model;

joey12300:
…about model_name_or_path in bigbird doc

### PR types

Bug fixes
### PR changes

Models
### Describe

add start training from checkpoint in bigbird model;add more detail about model_name_or_path in bigbird doc


wawltor (approved):
LGTM

PR #150, opened 2021-03-16
Fix num_steps usage to early return.

guoshengCS:
### PR types

Bug fixes
### PR changes

Models
### Describe

Fix num_steps usage to early return.


ZeyuChen (approved):
LGTM

PR #180, opened 2021-03-22
Fix load model for the pretrained path in bert 

wawltor:
### PR types
Bug fixes

### PR changes
Models 

### Describe
Fix load model for the pretrained path in bert


guoshengCS (approved):
LGTM

PR #213, opened 2021-04-01
extract->extractall

joey12300:
### PR types

Performance optimization
### PR changes

APIs
### Describe

1. use extractall instead of extract to optimize the decompression
2. add requirements


ZeyuChen (dismissed):
LGTM

PR #357, opened 2021-05-10
 Optimize CRF Perf 

joey12300:
### PR types

Performance optimization
### PR changes

Models
### Description

1. Optimize crf performance.
2. Fix ernie_crf negative label bug

### DIFF
--- a/paddlenlp/layers/crf.py
+++ b/paddlenlp/layers/crf.py
@@ -21,6 +21,13 @@ from paddlenlp.layers import sequence_mask
 __all__ = ['LinearChainCrf', 'LinearChainCrfLoss', 'ViterbiDecoder']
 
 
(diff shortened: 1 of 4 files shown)


paddlenlp/layers/crf.py
```diff
@@ -21,6 +21,13 @@
 __all__ = ['LinearChainCrf', 'LinearChainCrfLoss', 'ViterbiDecoder']
 
 
+def log_sum_exp(vec, dim=0):
+    # avoid underflow and overflow
```
wawltor: avoid -> Avoid

joey12300: done

wawltor (approved):
LGTM

PR #387, opened 2021-05-14
fix the bug in load_state_dict in paddlepaddle 2.1

wawltor:
### PR types
Bug fixes 

### PR changes
Models 

### Description
fix the bug in load_state_dict in paddlepaddle 2.1


ZeyuChen (approved):
LGTM

PR #393, opened 2021-05-17
optimize the print message for metric in glue task

wawltor:
### PR types
Function optimization 

### PR changes
Models

### Description
optimize the print message for metric in glue task


ZeyuChen (approved):
LGTM

PR #442, opened 2021-05-25
update returned data of cote dataset

Steffy-zxf:
### PR types

Others 

### PR changes

APIs

### Description

* return entity data from cote dataset


ZeyuChen (approved):
LGTM

PR #565, opened 2021-06-15
Change load_dataset design for loading local datasets

smallv0221:
### PR types

Function optimization
### PR changes

APIs 
### Description

Change load_dataset design for loading local datasets


guoshengCS (approved):
LGTM

PR #588, opened 2021-06-18
[GPT]. Optimize the speed of static graph, fix num_heads, add gpt2-small-en for CE.

ZHUI:
### PR types

Performance optimization
### PR changes

Others
### Description

Optimize the speed of static graph, fix num_heads, add gpt2-small-en for CE.


wawltor (approved):
LGTM

PR #631, opened 2021-06-25
Update run_glue.py

Grant-Rk:

### DIFF
--- a/examples/benchmark/glue/run_glue.py
+++ b/examples/benchmark/glue/run_glue.py
@@ -49,6 +49,7 @@ METRIC_CLASSES = {
     "mnli": Accuracy,
     "qnli": Accuracy,
     "rte": Accuracy,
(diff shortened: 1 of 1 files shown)


examples/benchmark/glue/run_glue.py
```diff
@@ -49,13 +49,15 @@
     "mnli": Accuracy,
     "qnli": Accuracy,
     "rte": Accuracy,
+    "chnsenticorp": Accuracy,
```
ZeyuChen, line 52: ChnSentiCorp is not a glue dataset?

ZeyuChen (commented):
ChnSentiCorp is not a glue dataset?

ZeyuChen:
It seems chnsenticorp is not glue dataset, so we will not merge this PR into run_glue.py. Thanks again!

PR #721, opened 2021-07-09
fix unified transformer dtype bug in win32

smallv0221:
### PR types

Bug fixes
### PR changes

APIs 
### Description

fix unified transformer dtype bug in win32


LiuChiachi (approved):
LGTM

PR #803, opened 2021-07-27
Simplify panda dependency in the taskflow API

wawltor:
### PR types
Function optimization

### PR changes
APIs 

### Description
remove the pandas dependency in the taskflow


ZeyuChen (approved):
LGTM

PR #818, opened 2021-07-29
fix waybill data_dir; remove data files

joey12300:
### PR types

Bug fixes
### PR changes

Models
### Description

fix waybill data_dir; remove data files


wawltor (approved):
LGTM

PR #834, opened 2021-08-03
fix the electra memory oom in the process of training

wawltor:
### PR types
Bug fixes 

### PR changes
Models

### Description
fix the electra oom in the process of training


ZeyuChen (approved):
LGTM

PR #863, opened 2021-08-09
optimize the time cost for the wordtag

wawltor:
### PR types
Function optimization 

### PR changes
Models

### Description
optimize the time cost for the wordtag


ZeyuChen (approved):
LGTM

PR #905, opened 2021-08-19
Transformer input id supports int32 & mv beam search v2

FrostML:
### PR types

Function optimization

### PR changes

Models

### Description

* Transformer input id supports int32 which comes from NPU `lookup_table_v2` operator doesn't support `int64`. 
* Move `beam_search_v2` to `InferTransformerModel`.  #838

PR #918, opened 2021-08-24
Fix beam search v2 dtype int32

FrostML:
### PR types

Bug fixes

### PR changes

Models

### Description

Fix beam search v2 dtype `int32`. Fix #905


guoshengCS (approved):
LGTM

PR #925, opened 2021-08-25
fix eval hang because of unique endpoint

smallv0221:
### PR types

Bug fixes
### PR changes

Others
### Description

fix eval hang because of unique endpoint


ZeyuChen (approved):
LGTM

PR #1005, opened 2021-09-10
Fix ernie csc decode

joey12300:
### PR types

Bug fixes 
### PR changes

APIs
#### Background
Some decode bugs show up when getting the wired inputs, just like below:
image

Because the number of tokens produced by tokenizer is different with the number of actual characters in some situaions, such as [UNK] or  English words show up. It needs to align the sequence of tokens and  the sequence of actual characters. There's so many situations which need to be aligned and it's hard to cover all situations.  So I decide to tokenize every character in sequence using list() instead of tokenizer.tokenize(). And then it is unnecessary to align,  so I remove inefficient aligning code in parse_decode().

#### Solution
1. Use list() to tokenize sequence
2. Remove inefficient aligning code in parse_decode

#### Now
image


ZeyuChen:
Please resolve conflicts. @joey12300

joey12300:
> Please resolve conflicts. @joey12300

done

ZeyuChen (approved):
LGTM

PR #1135, opened 2021-10-11
remove the unused load in the ner task of taskflow

wawltor:
### PR types
 Function optimization

### PR changes
 APIs
### Description
remove the unused load in the ner task of taskflow


ZeyuChen (approved):
LGTM

PR #1166, opened 2021-10-14
update the download check for the Taskflow

wawltor:
### PR types
Function optimization 

### PR changes
APIs
### Description
update the download check for the Taskflow


ZeyuChen (approved):
LGTM

PR #1229, opened 2021-10-25
[BUGFIX] Do broadcast right

gongel:
### PR types

Bug fixes
### PR changes

Others
### Description

- In `labels = np.where(labels != -100, labels, tokenizer.pad_token_id)`, `labels != -100` requires type of `labels` should be  `numpy.ndarray`
- In faster mode, `alpha==0` means no length_penalty and keeps same with normal mode.


guoshengCS (approved):
LGTM

PR #1235, opened 2021-10-26
patch only once

felixhao28:
### PR types
Others

### PR changes
Others

### Description
Improve the cmake process by avoiding duplicated patching, because if you run cmake multiple times, duplicated patching will corrupt the source files and makes the build to fail.

I will create an empty file named "paddle_cmake_patched" if patching succeeds. And check for this file's existence before performing another patching.


guoshengCS:
Thanks for your contribution!  We have updated the cmake file by #1122, and it seems no duplicated patching by testing . Could you please try the develop code or version after 2.1.  And to allow developers change the patches, sometimes we truly want to re-patch and then re-run FasterTransformer building when revising the patches.

felixhao28:
Thanks for letting me know. Since it is already covered in another PR, I am closing this one.

PR #1273, opened 2021-11-04
fix some conner case in text correction decode

joey12300:
### PR types

Bug fixes
### PR changes

APIs
### Description

fix some conner case in text correction decode


wawltor (approved):
LGTM

PR #1297, opened 2021-11-11
Add more args for generate API

smallv0221:
### PR types

New features
### PR changes

APIs
### Description

Add more args for generate API


gongel (approved):
LGTM

PR #1300, opened 2021-11-12
Fix FasterGeneration(FasterEncoder) when loading a lib without fusion_encoder to use FasterEncoder

FrostML:
### PR types

Bug fixes

### PR changes

APIs

### Description

Fix FasterGeneration(FasterEncoder) when load a lib without fusion_encoder to use FasterEncoder. 

After loading a custom op lib in `enable_faster_encoder()`, the original `TransformerEncoder`'s `forward` will be replaced by the faster version. 
However, if there is no `fusion_encoder` in the lib, the replacement cannot be undo.

### DIFF
--- a/paddlenlp/ops/faster_transformer/transformer/decoding.py
+++ b/paddlenlp/ops/faster_transformer/transformer/decoding.py
@@ -501,7 +501,10 @@ class InferTransformerDecoding(nn.Layer):
                 logger.warning(
                     "The specified decoding_lib does not exist, and it will be built automatically."
                 )
-            load("FasterTransformer", verbose=True)
+            load(
+                "FasterTransformer",
+                verbose=True,
+                opnames=["fusion_decoding", "fusion_force_decoding"])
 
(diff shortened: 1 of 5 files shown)


paddlenlp/ops/faster_transformer/transformer/decoding.py
```diff
@@ -501,7 +501,10 @@ def __init__(self,
                 logger.warning(
                     "The specified decoding_lib does not exist, and it will be built automatically."
                 )
-            load("FasterTransformer", verbose=True)
+            load(
+                "FasterTransformer",
+                verbose=True,
+                opnames=["fusion_decoding", "fusion_force_decoding"])
```
guoshengCS, line 507: 感觉不是特别好的设计，就是load时其实应该知道会加载哪些内容，不然也不会去load，而load的这些参数通常是要说明要去load哪些内容，如果费时间成功编完了又发现op没有load进来就感觉有点奇怪

PR #1321, opened 2021-11-17
Add amp training option for base ernie sequence classification task

joey12300:
### PR types

Function optimization
### PR changes

Models
### Description

Add amp training option for base ernie


wawltor (approved):
LGTM

PR #1422, opened 2021-12-09
Fix paddle.sum in FasterGPT

FrostML:
### PR types

Bug fixes

### PR changes

APIs

### Description

The returned type of `paddle.sum` is `bool` when input is `bool` in previous paddle.
Fix #1395


guoshengCS (approved):
LGTM

PR #1518, opened 2021-12-28
fix the ips of Bert in the static graph

Yanxing-Shi:
### PR types

Bug fixes

### PR changes

Models

### Description

There are two main problems for the error in the ips results of Bert:
- when static bert is trained by many cards, Each card will read a complete HDF5 file and it is not the same, but the `create_pretraining_dataset` is sharding again, causing some data to miss reading.
- There are problems with the ips calculation formula, the `total_samples` is where each `train_data_loader` is reset to 0, global_step is not. When `global_steps % args.logging_steps == 0`, the `total_samples` may not be the same each time.So the total time is the sum of reader and trainer, and I think the correct formula is `args.batch_size / (reader_cost_avg.get_average() + train_cost_avg.get_average())`  

After testing, the order of magnitude of ips is consistent with Bert of different framework, please review for confirmation.


guoshengCS (approved):
LGTM. 也请 @wawltor 再确认下

wawltor (approved):
LGTM

PR #1531, opened 2021-12-29
use suitable models for PP-MiniLM and SKEP

1649759610:
### Description

1. use PPMiniLMForSequenceClassification directly instead of loading it with Ernie Class.
2. use SKEPForXXX models in PaddleNLP, not implement it by personally.
3. refine coding with pre-commit.


tianxin1860 (approved):
LGTM

PR #1577, opened 2022-01-11
fix dtype problem when performing quantization function in windows

1649759610:
### Description

fix dtype problem when performing quantization function in windows


tianxin1860 (approved):
LGTM

PR #1599, opened 2022-01-17
Update shape range setting method

LiuChiachi:
### PR types

Plan optimization 
### PR changes

Models
### Description

Update shape range setting method,  and change 'collect_shape'  to 'set by users'


tianxin1860 (approved):
LGTM

PR #1656, opened 2022-01-28
Fix Generation expand_inputs_for_generation  when corresponding param is None

FrostML:
### PR types

Bug fixes

### PR changes

APIs

### Description

Fix Generation expand_inputs_for_generation  when corresponding param is None.


smallv0221 (approved):
LGTM

PR #1675, opened 2022-02-15
fix: bash typo

tao-hpu:
the bash is --logdir instead logdir.


ZeyuChen (approved):
LGTM, Thanks for your contributions

PR #1717, opened 2022-03-02
FasterUnifiedTransformer/PLATO support dy2sta

FrostML:
### PR types

Others

### PR changes

Models

### Description

FasterUnifiedTransformer/PLATO support dy2sta. 
Provide Paddle Inference python script. 

#### NOTE:
There is a problem when using Paddle Inference python API. 

That is: 
First, it is necessary to call `load_op_meta_info_and_register_op()` to load custom op lib before `create_predictor()`. 
However, in `create_predictor()`, the same load operation will be processed again which leads to user-confused logs as follow: 
image
If we do NOT call `load_op_meta_info_and_register_op()` first, then another exception which is `operator is not registered` will be throw. 

As for now, we can use `disable_glog_info()` to shut down all logs while the inference team may fix this within this year.

### DIFF
--- a/paddlenlp/ops/faster_transformer/transformer/decoding.py
+++ b/paddlenlp/ops/faster_transformer/transformer/decoding.py
@@ -1158,33 +1158,39 @@ class InferUnifiedDecoding(nn.Layer):
             "linear_bias": None
         }
         if self._use_fp16_decoding:
-            for mod in self._model.encoder.layers:
-                self.sub_modules["slf_q_weight"].append(
-                    paddle.concat(
-                        [
-                            transfer_param(
-                                mod.self_attn.q_proj.weight, restore_data=True),
-                            transfer_param(
+            for i, mod in enumerate(self._model.encoder.layers):
+                q_weight = paddle.concat(
+                    [
+                        transfer_param(
+                            mod.self_attn.q_proj.weight,
+                            restore_data=True), transfer_param(
                                 mod.self_attn.k_proj.weight, restore_data=True),
-                            transfer_param(
-                                mod.self_attn.v_proj.weight, restore_data=True)
-                        ],
-                        axis=-1))
-                self.sub_modules["slf_q_bias"].append(
-                    paddle.concat(
-                        [
-                            transfer_param(
-                                mod.self_attn.q_proj.bias,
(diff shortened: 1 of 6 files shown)


paddlenlp/ops/faster_transformer/transformer/faster_transformer.py
```diff
@@ -886,20 +885,32 @@ def generate_logits_mask(self, use_fp16_decoding):
             return logits_mask_t
 
     @staticmethod
-    def expand_inputs_for_generation(input_ids, expand_size, **model_kwargs):
+    def expand_inputs_for_generation(input_ids, expand_size, token_type_ids,
+                                     position_ids, attention_mask, seq_len,
+                                     role_ids, **model_kwargs):
```
guoshengCS, line 890: 动转静的时候除了forward以外其他方法里的参数也是必须要从model_kwargs里拿出来吗

guoshengCS (approved):

paddlenlp/ops/faster_transformer/transformer/decoding.py
```diff
@@ -1381,13 +1402,13 @@ def __init__(self,
...
-                "linear_weight"] = [self._model.lm_head.decoder_weight.t()]
+
+            self.dec_weight = self._model.lm_head.decoder_weight.t()
+            self.sub_modules["linear_weight"] = [self.dec_weight]
 
             # NOTE: Fix self._model.lm_head.decoder_bias been changed in FT.
-            self.sub_modules["linear_bias"] = [
-                paddle.assign(self._model.lm_head.decoder_bias)
-            ]
+            self.dec_bias = paddle.assign(self._model.lm_head.decoder_bias)
```
guoshengCS, line 1410: 这里需要assign还是因为之前那个未知的bias改变的原因吗

FrostML: 是的。

paddlenlp/ops/faster_transformer/transformer/faster_transformer.py
FrostML: 动转静解析不了 model_kwargs。

PR #1896, opened 2022-04-02
To avoid recursive import temporarily

guoshengCS:
### PR types

Bug fixes 
### PR changes

Others
### Description

To avoid recursive import temporarily。


LiuChiachi:
It seems that it doesn't work. 
b84343f5cf85db7e41a3d5707c444ab4



How about:
image


Because` model_utils.py` will be always imported.

guoshengCS:
> It seems that it doesn't work. b84343f5cf85db7e41a3d5707c444ab4
> 
> How about: image
> 
> Because` model_utils.py` will be always imported.

Done. Thanks

LiuChiachi (approved):
LGTM

jiweizhangxu:
66666

PR #2014, opened 2022-04-26
[ehealth] fix epoch problem when max_steps is not setted

LemonNoel:
### PR types

Bug fixes

### PR changes

Others

### Description

- Fix epoch problem when max_steps is not setted


ZHUI (approved):
LGTM

PR #2085, opened 2022-05-09
Add pre-trained model downloads logging

yingyibiao:
### PR types

Function optimization
### PR changes

Models
### Description

Logging pre-trained model downloads statistics.


linjieccc (approved):
LGTM

PR #2105, opened 2022-05-10
Fix return offsets mapping in python tokenizer

joey12300:
### PR types

Bug fixes
### PR changes

APIs
### Description

Fix return offsets mapping


ZeyuChen (approved):

smallv0221 (approved):
LGTM

PR #2114, opened 2022-05-11
Replace logger.warn with warnings.warn in tokenizer to avoid duplicated warning

guoshengCS:
### PR types

Function optimization 
### PR changes

APIs
### Description

Replace logger.warn with warnings.warn to avoid duplicated warning.


ZHUI (approved):
LGTM

PR #2195, opened 2022-05-17
fix unexpected terminate and save the correct program for pipeline

sandyhouse:
### PR types

Bug fixes
### PR changes

Others
### Description

1. avoid unexpected terminate by calling train_data_loader.reset()
2. use the correct program to save persistable for pipeline which is program._pipeline['section_program']


ZHUI (approved):
LGTM

PR #2223, opened 2022-05-18
[Bug fix] use executor for mp and mp+dp instead of parallel_executor

sandyhouse:
### PR types

Bug fixes
### PR changes

Others
### Description

[Bug fix] use executor for mp and mp+dp instead of parallel_executor


ZHUI (approved):
LGTM

PR #2448, opened 2022-06-08
Change cpu count func

yeliang2258:
### PR types

Others

### PR changes

Others

### Description

1. Change cpu count func from multiprocessing module to psutil module，use the number of CPU cores instead of the number of threads


ZeyuChen (approved):
LGTM

Issue #2520, opened 2022-06-15
model_config.json not exist after calling from_pretrained


wj-Mcat:
## Description

If I load the model with `from_pretrained` method and copy the cache_dir(`/root/.paddlenlp/models/<model-name>`) to other dist or dir(eg, /path/to/model). Then, I load it with `from_pretrained` method. In this time, it will failed because of that there is no `model_config.json` file. 

You can use the following code to reproduce the problem:
```python
import shutil, os
from paddlenlp.transformers.model_utils import MODEL_HOME
from paddlenlp.transformers.bert.modeling import BertForPretraining

model_name = 'bert-base-uncased'
cache_dir = os.path.join(MODEL_HOME, model_name)
# keep the cache dir clean
shutil.rmtree(cache_dir, ignore_errors=True)

print('downloading bert weight file from server ...')
modle = BertForPretraining.from_pretrained(model_name)

print('loading bert weight from local dir ...')
BertForPretraining.from_pretrained(cache_dir)
```

the error trackstack info:

```shell
downloading bert weight file from server ...
[2022-06-15 06:30:01,097] [    INFO] - Downloading https://bj.bcebos.com/paddlenlp/models/transformers/bert-base-uncased.pdparams and saved to /root/.paddlenlp/models/bert-base-uncased
[2022-06-15 06:30:01,097] [    INFO] - Downloading bert-base-uncased.pdparams from https://bj.bcebos.com/paddlenlp/models/transformers/bert-base-uncased.pdparams
100%|███████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 775M/775M [00:11<00:00, 72.2MB/s]
W0615 06:30:12.412072 25024 gpu_context.cc:278] Please NOTE: device: 0, GPU Compute Capability: 7.0, Driver API Version: 10.2, Runtime API Version: 10.2
W0615 06:30:12.417645 25024 gpu_context.cc:306] device: 0, cuDNN Version: 7.6.
loading bert weight from local dir ...
[2022-06-15 06:30:36,430] [    INFO] - Downloading /root/.paddlenlp/models/bert-base-uncased/model_state.pdparams and saved to /root/.paddlenlp/models/bert-base-uncased
Traceback (most recent call last):
  File "a.py", line 14, in <module>
    BertForPretraining.from_pretrained(cache_dir)
[...]

PR #2521, opened 2022-06-15
Improve tokenizer

wj-Mcat:
### PR types
Bug fixes

### PR changes
APIs

### Description
try to fix issue: https://github.com/PaddlePaddle/PaddleNLP/issues/2520

I believe there are some more potential bugs in `from_pretrained`, so more boundary tests are required, I have add the basic one.


paddlenlp/transformers/model_utils.py
```diff
@@ -203,19 +203,19 @@ def from_pretrained(cls, pretrained_model_name_or_path, *args, **kwargs):
...
                 model = BertForSequenceClassification.from_pretrained('./my_bert/')
         """
-        pretrained_models = list(cls.pretrained_init_configuration.keys())
         resource_files = {}
         init_configuration = {}
         load_state_as_np = kwargs.pop("load_state_as_np", False)
 
         # From built-in pretrained models
-        if pretrained_model_name_or_path in pretrained_models:
+        if pretrained_model_name_or_path in cls.pretrained_init_configuration:
```
wj-Mcat, line 211: there are some grammar sugar in python:

```python
'key' in dict_data
```

is equal is to :

```python
'key' in list(dict_data.keys())
```

And there are more improvements:

* time complexity: O(n) -> O(1)
* memory complexity: O(n) -> O(0)

So I change this code.

paddlenlp/transformers/model_utils.py
```diff
@@ -270,6 +270,16 @@ def from_pretrained(cls, pretrained_model_name_or_path, *args, **kwargs):
                 init_kwargs = json.load(f)
         else:
             init_kwargs = init_configuration
+
+            model_config_file_path = os.path.join(default_root,
+                                                  cls.model_config_file)
+            # check if there is model config file in cache directory
+            if pretrained_model_name_or_path in cls.pretrained_init_configuration and init_kwargs is not None and not os.path.exists(
```
wj-Mcat: I have add the checking situation:

```python
init_kwargs is not None
````

So, I'm curious about that whether model_config_file and init_kwargs will be None at the same time ? I think this will be boundary case which can be avoided by developers although.

File: tests/transformers/test_tokenizer.py [hunk 3]
```diff
@@ -0,0 +1,95 @@
...
+        shutil.rmtree(cache_dir, ignore_errors=True)
+
+        model: PretrainedModelClass = PretrainedModelClass.from_pretrained(
+            model_name)
+        self.assertTrue(
+            os.path.exists(os.path.join(cache_dir, model.model_config_file)))
+
+        # TODO: make this test code pass
+        # from pretrained from the dir
+        # PretrainedModelClass.from_pretrained(cache_dir)
```
wj-Mcat, line 118: this code will not pass the unit test.

wj-Mcat (commented):
Here are some considerations in my side, let's discuss about them.  @guoshengCS

wj-Mcat:
Here are some testing status about the above unit test:

image

So, there may be some other problems in large pretrained model.

wj-Mcat:
## Introduction

After making more tests under tokenizer, I found that there are some more warnings which will make unit test failed.  the error infos are:

```shell
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/root/project/paddle/PaddleNLP/tests/common_test.py", line 68, in wrapper
    self.assertFalse(len(w) > 0, msg)
AssertionError: True is not false : unclosed file <_io.TextIOWrapper name='/root/.paddlenlp/models/gpt2-large-en/gpt-en-vocab.json' mode='r' encoding='UTF-8'>

======================================================================
FAIL: test_pretrained_tokenizer_18_gpt2_medium_en (tests.transformers.test_tokenizer.TestPretrainedFromPretrained)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/root/project/paddle/PaddleNLP/tests/common_test.py", line 68, in wrapper
    self.assertFalse(len(w) > 0, msg)
AssertionError: True is not false : unclosed file <_io.TextIOWrapper name='/root/.paddlenlp/models/gpt2-medium-en/gpt-en-vocab.json' mode='r' encoding='UTF-8'>

======================================================================
FAIL: test_pretrained_tokenizer_19_gpt2_en (tests.transformers.test_tokenizer.TestPretrainedFromPretrained)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/root/project/paddle/PaddleNLP/tests/common_test.py", line 68, in wrapper
    self.assertFalse(len(w) > 0, msg)
AssertionError: True is not false : unclosed file <_io.TextIOWrapper name='/root/.paddlenlp/models/gpt2-en/gpt-en-vocab.json' mode='r' encoding='UTF-8'>

======================================================================
FAIL: test_pretrained_tokenizer_20_gpt2_small_en (tests.transformers.test_tokenizer.TestPretrainedFromPretrained)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/root/project/paddle/PaddleNLP/tests/common_test.py", line 68, in wrapper
    self.assertFalse(len(w) > 0, msg)
AssertionError: True is not false : unclosed file <_io.TextIOWrapper name='/root/.paddlenlp/models/gpt2-small-en/gpt-en-vocab.json' mode='r' encoding='UTF-8'>

----------------------------------------------------------------------
```

## Analysis

According to the error info, I find that there are many file handlers are not closed after opening.  So I find the key code:

```python
# https://github.com/PaddlePaddle/PaddleNLP/blob/develop/paddlenlp/transformers/gpt/tokenizer.py#L392
self.encoder = json.load(open(vocab_file))
bpe_data = open(merges_file, encoding='utf-8').read().split('\n')[1:-1]
```

in this code, it don't close the file handler and the python standard library will log warning which will make unit test fail with the following code:

```python
# https://github.com/PaddlePaddle/PaddleNLP/blob/develop/tests/common_test.py#L59
def wrapper(self, *args, **kwargs):
    with warnings.catch_warnings(record=True) as w:
        warnings.resetwarnings()
        # ignore specified warnings
        warning_white_list = [UserWarning]
        for warning in warning_white_list:
            warnings.simplefilter("ignore", warning)
        func(self, *args, **kwargs)
        msg = None if len(w) == 0 else w[0].message
        self.assertFalse(len(w) > 0, msg)

return wrapper
```

the suggested style is using with keyword to handle the file stream operation.

wj-Mcat:
After using the with style to handle file, all of unit test pass.

image

Should I add the improvement in this pr ?

paddlenlp/transformers/model_utils.py
guoshengCS: No problem.

paddlenlp/transformers/model_utils.py
```diff
@@ -270,6 +270,16 @@ def from_pretrained(cls, pretrained_model_name_or_path, *args, **kwargs):
...
+                    model_config_file_path):
+                with io.open(model_config_file_path, 'w',
+                             encoding='utf-8') as f:
+                    json.dump(init_kwargs, f)
```
guoshengCS: Maybe multi-processing need to be tested for concurrent writing which is common for training code.

guoshengCS: Maybe multi-processing need to be tested for concurrent writing which is common for training code.

paddlenlp/transformers/tokenizer_utils_base.py
```diff
@@ -1690,6 +1689,10 @@ def convert_added_tokens(obj):
                 "Special tokens have been added in the vocabulary, make sure the associated word embeddings are fine-tuned or trained."
             )
 
+        # save all of related things into default root dir
+        if pretrained_model_name_or_path in cls.pretrained_init_configuration:
+            tokenizer.save_pretrained(default_root)
```
guoshengCS, line 1702: Should this only be triggered only when using `init_configuration` and not exist, just like that in `model.from_pretrained`

guoshengCS, line 1702: Should this be triggered only when using `init_configuration` and not exist, just like that in `model.from_pretrained`

guoshengCS (commented):
Considering the extra complexity when taking multi-processing into account, should we use a simpler solution which only outputs logs to remind users to save first.

guoshengCS:
It seems there are some expired commits
image

paddlenlp/transformers/model_utils.py
wj-Mcat: I have pushed the unit test for `multi-process` mode to `from_pretrained` the model. Here are some test result on latest code:

* code

```python

class FakePretrainedModel(PretrainedModel):

    def __init__(self):
        super().__init__()
        self.linear = nn.Linear(2, 2)

    def forward(self, *args, **kwargs):
        pass


class TestPretrainedFromPretrained(CpuCommonTest):
    """module for test pretrained model"""

    def setUp(self):
        self.temp_dir = TemporaryDirectory()
        model = FakePretrainedModel()
        model.save_pretrained(self.temp_dir.name)

    def do_pretrained_in_process(self):
        FakePretrainedModel.from_pretrained(self.temp_dir.name)

    @parameterized.expand([(1, ), (8, ), (20, ), (50, ), (100, ), (1000, )])
    def test_model_config_writing(self, process_num: int):
        for _ in range(process_num):
            process = Process(target=self.do_pretrained_in_process)
            process.start()
```

* result

image

wj-Mcat: tested under 1000 processes to write the target file at same time, there are no bugs.

paddlenlp/transformers/tokenizer_utils_base.py
wj-Mcat: there are many resource files in tokenizer, so I think `checking if the target file exist` will be in `save_pretrained` dir.

Another question, If user modify the tokenizer/model related configuration file, will it run with no issue ?  or , are there some situations that the developer should modify the configuration to suit their model ???

wj-Mcat:
When I pull the latest code from remote master branch or update branch in github pr webpage, it will create a new commit in this pr. But if this pr is merged, it will not appear in the develop history commits.

wj-Mcat:
ping @guoshengCS  is there anything to be updated ?

paddlenlp/transformers/tokenizer_utils_base.py
guoshengCS: Sure, `Configuration` is another issue we need to resolve.

guoshengCS (dismissed):
LGTM

paddlenlp/transformers/model_utils.py
guoshengCS: Need to use `save_model_config` to save, which differs with dumping directly.

wj-Mcat: Yes, get it done.

guoshengCS (approved):

guoshengCS:
> When I pull the latest code from remote master branch or update branch in github pr webpage, it will create a new commit in this pr. But if this pr is merged, it will not appear in the develop history commits.

Sure it is, however these expired commit messages would also be merged in which results to redundant messages, and that is what we want to avoid.

tests/transformers/test_tokenizer.py
```diff
@@ -0,0 +1,147 @@
...
+class TestPretrainedFromPretrained(CpuCommonTest):
+    """module for test pretrained model"""
+
+    def setUp(self):
+        self.temp_dir = TemporaryDirectory()
+        model = FakePretrainedModel()
+        model.save_pretrained(self.temp_dir.name)
+
+    def do_pretrained_in_process(self):
+        FakePretrainedModel.from_pretrained(self.temp_dir.name)
```
guoshengCS, line 96: Need to test multi-processing of tokenizer since `PretrainedModel` use check before saving while `PretrainedTokenizer` not.

wj-Mcat: Yes, there are also some improvements to do:
- [ ] test `PretrainedTokenizer` under multi-processing
- [ ] change  multi-processing under multi-stage k(0<k<=8)  processors which will be more stable for the testing. 

All of above improvements will be done in the next few days.

PR #2544, opened 2022-06-16
Update trans_to_json.py

ArtificialZeng:
Bug fix：如果不加，win报如下错误，加进去就好了。
UnicodeDecodeError: 'gbk' codec can't decode byte 0x80 in position 45: illegal multibyte sequence


ZeyuChen (approved):
LGTM. Thanks for your contributions!

PR #2620, opened 2022-06-23
test pretrained tokenizer & model with multi process

wj-Mcat:
### PR types
Performance optimization

### PR changes
Models

### Description
try to test tokenizer & model with process pool to test the stability of Pretrained tokenizer & model.


wj-Mcat:
ping @guoshengCS

wj-Mcat:
will the content of this pr be by the new unit test which is post by yibiao? If so, I think this pr should be closed.  @guoshengCS

PR #2669, opened 2022-06-28
optimize moe with fused op

FeixLiu:
### PR types
Others

### PR changes
Others

### Description
optimize moe


FeixLiu:
Since the moe benchmark cases will use nearly all memory of gpu, the extra memory occupation introduced by fused op will bring OOM to all cases, just close this pr.

PR #2732, opened 2022-07-05
fix error when relation is empty

littletomatodonkey:
### PR types
Bug fixes

### PR changes
Models

### Description
fix error when relation is empty(`{}`)


yingyibiao (approved):
LGTM

PR #2875, opened 2022-07-26
add func get_special_tokens_mask()

zhiyongLiu1114:
add func get_special_tokens_mask() instead of using that in superclass which will lead to conflict with func build_inputs_with_special_tokens()


guoshengCS (approved):
LGTM.

guoshengCS:
Thanks for your contribution!

guoshengCS:
resolve #2915

PR #2910, opened 2022-07-28
Fix FT sampling performance issue

FrostML:
### PR types

Bug fixes

### PR changes

Others

### Description

When using sampling strategy in FT, the curand initialization function's performance's instability may cause the performance change sharply. This PR tries to fix this problem.


guoshengCS (commented):
如果验证没问题的话就合入吧

guoshengCS (approved):

PR #2954, opened 2022-08-03
Fix random generate question ui error and  search results re display of pipelines

w5688414:
### PR types

+ Bug fixes

https://github.com/PaddlePaddle/PaddleNLP/issues/2949

### PR changes

+ APIs

### Description

+ Fix ui search results re display bugs
+ Fix random generate question ui


tianxin1860 (approved):
LGTM

PR #2977, opened 2022-08-06
remove duplicate code in valid text input

BasicCoder:
only need do once valid

### PR types

Function optimization
### PR changes

Others
### Description

        if not _is_valid_text_input(text):
            raise ValueError(
                "text input must of type `str` (single example), `List[str]` (batch or single pretokenized example) "
                "or `List[List[str]]` (batch of pretokenized examples).")

        if not _is_valid_text_input(text):
            raise ValueError(
                "text input must of type `str` (single example), `List[str]` (batch or single pretokenized example) "
                "or `List[List[str]]` (batch of pretokenized examples).")

only need to do once


ZeyuChen (approved):
LGTM, thanks for your contributions.

PR #3052, opened 2022-08-15
[Benchmark] fix random rank in pdc

sljlp:
### PR types

Bug fixes
### PR changes

Others
### Description

In PDC, the node ranks may be randomly distributed, which makes we confused about which node is the last one


ZHUI (approved):
LGTM

PR #3101, opened 2022-08-22
[GPT] Update default lr of args.py

ZHUI:
### PR types

Bug fixes
### PR changes

Others
### Description

Update default lr of args.py


LemonNoel (approved):
LGTM

PR #3215, opened 2022-09-07
Remove boost library.

joey12300:
### PR types

Performance optimization
### PR changes

Others
### Description

Remove boost library dependency. Decrease the size of faster_tokenizer cpp library from 200M to 30M


ZeyuChen (approved):
可以考虑发个0.2版本

PR #3218, opened 2022-09-07
Fix default device

gongel:
### PR types

Bug fixes
### PR changes

APIs
### Description

- Taskflow of text summarization  device defaults to gpu, it is not good. Fix default device to environment device.


wawltor (approved):
LGTM

PR #3411, opened 2022-10-08
Support past_key_values argument for Electra

sijunhe:
### PR types
Function optimization

### PR changes
APIs

### Description
Addressing `past_key_values` part of #3382
Adding `past_key_values` argument for Electra models to support use cases such as prefix-tuning.
Also adds an unit test that checks the new arguments (Should create separate PRs to add this test to all models that supports `past_key_values`)


paddlenlp/transformers/electra/modeling.py
```diff
@@ -585,6 +592,18 @@ def forward(self,
                 When the data type is float, the `masked` tokens have `-INF` values and the others have `0` values.
                 It is a tensor with shape broadcasted to `[batch_size, num_attention_heads, sequence_length, sequence_length]`.
                 Defaults to `None`, which means nothing needed to be prevented attention to.
+            past_key_values (tuple(paddle.nn.MultiHeadAttention.Cache), optional):
```
guoshengCS: To be friendly to new comers, we prefer to be consistent with widely used HF transformers. For `past_key_values`, maybe we can use the same data type as HF transformers, which is:

```
past_key_values (`tuple(tuple(torch.FloatTensor))` of length `config.n_layers` with each tuple having 4 tensors of shape `(batch_size, num_heads, sequence_length - 1, embed_size_per_head)`)
```

While `MultiHeadAttention` uses `paddle.nn.MultiHeadAttention.Cache` as the data type, we can actually add a wrapper or hook converting between `tuple(Tensor)` and `MultiHeadAttention.Cache`, and BERT/ERNIE uses [monkey patches on `paddle.nn.TransformerEncoderLayer`](https://github.com/PaddlePaddle/PaddleNLP/blob/develop/paddlenlp/transformers/model_outputs.py#L168) to achieve this.  

What do you think about it?

sijunhe: Addressed.
One minor thing that is different from HF is that here we have 2 tensors in `past_key_values` instead of 4 tensors (key, value, cross attention key, cross attention values). Since electra is encoder-only, there is no cross attention key and values. This way we can directly use `Cache` object and don't need to bother with the `StaticCache` object

tests/transformers/electra/test_modeling.py
```diff
@@ -133,6 +133,60 @@ def create_and_check_electra_model(
...
+                outputs_without_cache.last_hidden_state.shape)
+            self.parent.assertFalse(
+                paddle.allclose(outputs_with_cache.last_hidden_state,
+                                outputs_without_cache.last_hidden_state))
+        else:
+            outputs_with_cache, _ = outputs_with_cache
+            self.parent.assertEqual(outputs_with_cache.shape,
+                                    outputs_without_cache.shape)
+            self.parent.assertFalse(
+                paddle.allclose(outputs_with_cache, outputs_without_cache))
```
guoshengCS, line 188: Maybe a more strict test is to guarantee values of output cache (when using past_key_values as input) and cache slice (when using a longer input_ids and without the past_key_values input) are same,  while the current test is okay.

guoshengCS (approved):
LGTM

PR #3598, opened 2022-10-28
Moe use adamw from paddle.optimizer

sljlp:
### PR types

Bug fixes
### PR changes

APIs
### Description

In MoE Model, the AdamW is defined by user before. Now we changed it to directly using paddle.optimizer.AdamW


ZHUI (approved):
LGTM

PR #3646, opened 2022-11-02
fix bug about template 3

westfish:
### PR types

Bug fixes

### Description

修复bug
[Bug]: [question generation] template 3 NameError: name 'tokenizer' is not defined #3634


wawltor (approved):
LGTM

PR #3755, opened 2022-11-14
Add LitEma for SD

firestonelib:
### PR types

New features
### PR changes

Others
### Description

Add LitEma for SD，use add_ op instead of torch sub_ op.


JunnYu (approved):
LGTM

PR #3803, opened 2022-11-17
bump evaluate to `0.3.0`

lvwerra:
### PR types
Others

### PR changes
Others

### Description
There is an issue with `evaluate` before 0.3.0 concerning the Hugging Face hub integration. I suggest upgrading to `0.3.0` which should not have any breaking changes and be functionally equivalent since this version might break in a few months.


sijunhe:
Thanks for the PR! @lvwerra and sorry about the late response.
@westfish would you plz verify that the PR doesn't break the question generation example?

PR #3924, opened 2022-11-28
[prompt] fix bug when no mask token in template

LemonNoel:
### PR types

Bug fixes

### PR changes

APIs

### Description

- Fix KeyError when there is not `{'mask'}` token defined in template


wawltor (approved):
LGTM

PR #3956, opened 2022-11-30
fix the client requeset for the serving

wawltor:
### PR types
Bug fixes 

### PR changes
Others

### Description
fix the client requeset for the serving


linjieccc (approved):
LGTM

PR #3960, opened 2022-11-30
[trainer] fix bug when batch size=1

LemonNoel:
### PR types

Bug fixes

### PR changes

APIs

### Description

- Add type checking in prediction_loop to avoid unexpected `logits[0]` when batch_size=1, which might lead to `Out of memory error`.

### DIFF
--- a/paddlenlp/trainer/trainer.py
+++ b/paddlenlp/trainer/trainer.py
@@ -30,53 +30,27 @@ import types
 from collections.abc import Mapping
 from pathlib import Path
 from typing import Any, Callable, Dict, List, Optional, Tuple, Union
-from packaging import version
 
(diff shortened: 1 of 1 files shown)


paddlenlp/trainer/trainer.py
```diff
@@ -1952,19 +1797,11 @@ def prediction_step(
             return (loss, None, None)
 
         logits = nested_detach(logits)
-        if len(logits) == 1:
+        if isinstance(logits, (list, tuple)) and len(logits) == 1:
```
LemonNoel, line 1800: 这里增加了类型检查，其他修改均为代码风格检查。

ZHUI (approved):
LGTM

PR #4021, opened 2022-12-06
[Prompt] PromptModelForSequenceClassification and unit tests

sijunhe:
### PR types
Function optimization

### PR changes
APIs

### Description
- Modify API of `PromptModelForSequenceClassification` to behave similarly to `AutoModelForSequenceClassification`
- Add unit tests


paddlenlp/prompt/prompt_model.py
```diff
@@ -0,0 +1,119 @@
...
+
+        if not return_dict:
+            output = (logits, hidden_states)
+            return ((loss,) + output) if loss is not None else output
+
+        return SequenceClassifierOutput(
+            loss=loss,
+            logits=logits,
+            hidden_states=hidden_states,
+        )
```
sijunhe: same logic as `AutoModelForSequenceClassification`

paddlenlp/prompt/prompt_trainer.py
```diff
@@ -43,10 +40,10 @@ class PromptTrainer(Trainer):
 
     def __init__(
         self,
-        model: Union[nn.Layer],
+        model: nn.Layer,
         tokenizer: PretrainedTokenizer,
-        criterion: Union[nn.Layer],
-        args: PromptTuningArguments = None,
+        criterion: Optional[nn.Layer] = None,
```
sijunhe, line 45: similar behavior as `Trainer`

tests/prompt/test_prompt_model.py
```diff
@@ -0,0 +1,67 @@
...
+    PromptDataCollatorWithPadding,
+    PromptModelForSequenceClassification,
+    SoftVerbalizer,
+)
+from paddlenlp.transformers import AutoModelForMaskedLM, AutoTokenizer
+
+
+class TestPromptModel(unittest.TestCase):
+    @classmethod
+    def setUpClass(cls):
```
LemonNoel, line 32: 建议新增一个单测，模型输入 `model: AutoModelForSequenceClassification`, `verbalizer: None`，保证类似 [EFL](https://github.com/PaddlePaddle/PaddleNLP/blob/develop/examples/few_shot/efl/run_train.py) 的算法能正常执行。

sijunhe: 已加入

tests/prompt/test_prompt_model.py
```diff
@@ -0,0 +1,87 @@
...
+        cls.label_words = {0: "0", 1: "1", 2: "2"}
+        cls.verbalizer = SoftVerbalizer(cls.label_words, cls.tokenizer, cls.model)
+        cls.data_collator = PromptDataCollatorWithPadding(cls.tokenizer, padding=True, return_tensors="pd")
+        cls.prompt_model = PromptModelForSequenceClassification(cls.model, cls.template, cls.verbalizer)
+
+    def test_sequence_classification_no_labels(self):
+        examples = [{"text": "百度飞桨深度学习框架"}, {"text": "这是一个测试"}]
+        encoded_examples = [self.template(i) for i in examples]
+        logits, hidden_states = self.prompt_model(**self.data_collator(encoded_examples))
+        self.assertTrue(logits.shape[0], len(examples))
```
LemonNoel: It seems that `assertTrue` only works for boolean type.

tests/prompt/test_prompt_model.py
```diff
@@ -0,0 +1,87 @@
+# Copyright (c) 2022 PaddlePaddle Authors. All Rights Reserved.
```
LemonNoel, line 1: You can add one line in Makefile to run the unittest.

```
unit-test:
	# only enable bert-test: there are many failed tests
	PYTHONPATH=$(shell pwd) pytest tests/transformers/bert
	PYTHONPATH=$(shell pwd) pytest tests/prompt
```

LemonNoel (changes requested):

paddlenlp/prompt/prompt_model.py
```diff
@@ -0,0 +1,119 @@
...
+            model_inputs.pop("masked_positions")
+        hidden_states = self.plm(**model_inputs)
+        if self.verbalizer is not None:
+            logits = self.verbalizer.process_outputs(hidden_states, input_dict["masked_positions"])
+        else:
+            logits = hidden_states
+
+        loss = None
+        if labels is not None:
+            if self.verbalizer is None:
```
LemonNoel: Directly raising Error when self.verbalizer is None will make `PromptModelForSequenceClassification` not applicable for EFL in `PromptTrainer`, since labels are always passed to `PromptModelForSequenceClassification`.

Maybe the condition can be modified as something like

```
if self.verbalizer is None and not isinstance(self.plm, xxForSeuquenceClassification)
```

LemonNoel (changes requested):

paddlenlp/prompt/prompt_model.py
LemonNoel: Or type checking could be added in initialization and it makes sure that self.plm is either `xxForSequenceClassification` or ``xxForMaskedLM``. Then even if self.verbalizer is None, the loss can also be calculated here with CE Loss.

paddlenlp/prompt/prompt_trainer.py
```diff
@@ -64,6 +61,9 @@ def __init__(
         if data_collator is None:
             data_collator = PromptDataCollatorWithPadding(tokenizer, padding=True, return_tensors="pd")
 
+        if criterion is None and (args.use_rgl or args.use_rdrop):
+            raise Exception("'To use 'use_rgl', use_rdrop', 'criterion' must be specified")
```
LemonNoel: It seems that the first single quotation mark is in the wrong place.

tests/prompt/test_prompt_model.py
```diff
@@ -0,0 +1,106 @@
...
+    SoftVerbalizer,
+)
+from paddlenlp.transformers import (
+    AutoModelForMaskedLM,
+    AutoModelForSequenceClassification,
+    AutoTokenizer,
+)
+
+
+class TestPromptModel(unittest.TestCase):
```
LemonNoel: According to @wj-Mcat , the class name of TestCase follows the rule: {name}Test, you can refer to: [transformers unittest](https://github.com/PaddlePaddle/PaddleNLP/blob/develop/tests/transformers/roberta/test_modeling.py#L399).

LemonNoel (changes requested):

LemonNoel (approved):
LGTM

PR #4028, opened 2022-12-06
[prompt] add unittest for template apis

LemonNoel:
### PR types

Others

### PR changes

Others

### Description

- Add UnitTest for ManualTemplate, SoftTemplate, PrefixTemplate and AutoTemplate
- Fix bugs which cause unexpected outputs.


paddlenlp/prompt/prompt_tokenizer.py
```diff
@@ -122,27 +131,41 @@ def _create_max_lengths_from_do_truncate(self, part_text: List[str], part_do_tru
...
+                    max_length -= len(part)
+                    has_short = True
+        assert max_length >= 0
+        avg_max_length = max_length // sum(do_truncate)
+        for index in range(len(part_text)):
+            if do_truncate[index] == 1:
+                max_lengths[index] = min(avg_max_length, max_length)
+                max_length -= max_lengths[index]
+                if max_length < 0:
+                    raise ValueError("Unexpected results")
```
wj-Mcat: It will be more readable if there are some more comments for this code block.  How do you think about ?

paddlenlp/prompt/prompt_tokenizer.py
```diff
@@ -122,27 +131,41 @@ def _create_max_lengths_from_do_truncate(self, part_text: List[str], part_do_tru
...
+        while has_short:
+            has_short = False
+            avg_max_length = max_length // sum(do_truncate)
+            for index, part in enumerate(part_text):
+                if do_truncate[index] == 1 and len(part) <= avg_max_length:
+                    do_truncate[index] = 0
+                    max_lengths[index] = len(part)
+                    max_length -= len(part)
+                    has_short = True
+        assert max_length >= 0
```
wj-Mcat: you can add a error message for this assert: `assert expression, "error message is here"`

paddlenlp/prompt/template.py
```diff
@@ -78,14 +78,17 @@ def __init__(self, prompt: str, tokenizer: PretrainedTokenizer, max_length: int,
...
 
     @prompt.setter
-    def prompt(self, prompt: str):
+    def prompt(self, prompt):
+        logger.warning("Prompt can not be modified once set.")
+
+    def set_prompt(self, prompt: str):
         if prompt is not None:
             if isinstance(prompt, str):
                 self._prompt = self.parse_template_string(prompt)
```
wj-Mcat, line 94: you can use prompt.setter method to handle `set_prompt` logic code. It will be more concise, do you think so?

```python
@prompt.setter
def prompt(self, prompt):
    if self._prompt is not None:
        logger.warning("Prompt can not be modified once set.")
        return 
    # code block in `set_prompt` method
    ...
```

tests/prompt/test_template.py
```diff
@@ -0,0 +1,202 @@
...
+
+import numpy as np
+import paddle
+from parameterized import parameterized
+
+from paddlenlp.prompt import AutoTemplate, ManualTemplate, PrefixTemplate, SoftTemplate
+from paddlenlp.transformers import AutoModelForMaskedLM, AutoTokenizer
+
+
+class TestTemplate(unittest.TestCase):
```
wj-Mcat: The class name of TestCase follows the rule: `{name}Test`, you can refer to: [transformers unittest](https://github.com/PaddlePaddle/PaddleNLP/blob/develop/tests/transformers/roberta/test_modeling.py#L399).

wj-Mcat (changes requested):
To make your unittest run in the lint workflow, you can add one line code to Makefile: 

```makefile
unit-test:
	# only enable bert-test: there are many failed tests
	PYTHONPATH=$(shell pwd) pytest tests/transformers/bert
	PYTHONPATH=$(shell pwd) pytest tests/prompt
```

so your test will be run in github action workflow temporarily, we will refactor test later.

paddlenlp/prompt/template.py
LemonNoel: This will cause Error in subclass, where prompt.setter is repeatedly called.

paddlenlp/prompt/prompt_tokenizer.py
LemonNoel: Comments are added in https://github.com/PaddlePaddle/PaddleNLP/pull/4028/commits/1ea904b6204accd5981880ec9f560f0a671953b5

paddlenlp/prompt/prompt_tokenizer.py
LemonNoel: Error message added in https://github.com/PaddlePaddle/PaddleNLP/pull/4028/commits/1ea904b6204accd5981880ec9f560f0a671953b5

tests/prompt/test_template.py
LemonNoel: Renamed as TemplateTest in https://github.com/PaddlePaddle/PaddleNLP/pull/4028/commits/8907013ca11532a3b5ba75c45ae34a59d70244a4

paddlenlp/prompt/template.py
wj-Mcat: If there is no error in the base-class, there should not be errors existing in subclass.  I think using one method to handle the `setter` of prompt will be more concise. If there is a technical problem that is difficult to solve, the current solution looks good to me.

wj-Mcat:
It seems that your test failed in test workflow.

LemonNoel:
> It seems that your test failed in test workflow.

It is fixed now.

wj-Mcat (approved):
LGTM

PR #4133, opened 2022-12-15
fix qg export and infer error in new paddle version

westfish:
### PR types

Bug fixes

### PR changes

Models

### Description

修复 unimo-text采用fp16预测报错 #4106


westfish:
close

PR #4154, opened 2022-12-19
Translate PaddleNLP/model_zoo/ernie-3.0/README_CN.md

HCQ14:
Translate PaddleNLP/model_zoo/ernie-3.0/README_CN.md


sijunhe (changes requested):
Thanks for your contribution, we really appreciate the help in improving our english docs!
However, the english readme looks machine-translated and many of the technical terms are not translated properly. 
We already wrote something similar in another repo. Would you mind copy some of the content from there? https://github.com/PaddlePaddle/models/blob/release/2.4/modelcenter/ERNIE-3.0/introduction_en.ipynb

PR #4192, opened 2022-12-21
Fix 0 dim tensor

FrostML:
### PR types

Bug fixes 

### PR changes

Models

### Description

Fix 0 dim tensor.


sijunhe (approved):
lgtm. I confirm that test is passing after this fix.

PR #4236, opened 2022-12-26
[CI] Speed up Ernie Gram test by using a tiny model config

sijunhe:
### Description
Currently ErnieGram model tests is slow and frequently fails in CI due to the large model size. Switching to a small model for CI.


wj-Mcat (approved):
LGTM

PR #4262, opened 2022-12-28
Update trainer.md

ymyjl:
Add skip_memory_metrics parameter

### PR types
 Others 

### PR changes
 Docs 

### Description
Add skip_memory_metrics which is used for whether to trace memory metrics


ZHUI (approved):
LGTM

PR #4302, opened 2022-12-30
fix running time for skep and uie

1649759610:
### PR types

Others

### PR changes

Docs

### Description

fix running time for skep and uie


wawltor (approved):
LGTM

PR #4346, opened 2023-01-05
[MoE] Remove Using Sharding API in legacy mode for MoE model

haohongxiang:
### PR types

Bug fixes
### PR changes

Others
### Description

[MoE] Remove Using Sharding API in legacy mode for MoE model


ZHUI (approved):
LGTM

PR #4462, opened 2023-01-12
Update group code and and unimo-text metric

chenxiaozeng:
### PR types

Update group code and and unimo-text metric.

### PR changes

Docs
### Description

Update group code and and unimo-text metric.


wawltor (approved):
LGTM

PR #4499, opened 2023-01-17
[Models] Save tokenizer config 

joey12300:
### PR types

Function optimization
### PR changes

Models
### Description

Save tokenizer config in export model directory


wawltor (approved):
LGTM

PR #4538, opened 2023-01-29
Temporary disable isort due to upstream bug

sijunhe:
### Description
Turning off isort temporary due to https://github.com/PyCQA/isort/issues/2077
As of 2023.1.29, bumping isort to 5.12.0 would resolve the error but requires python3.8+
Track the issue: https://github.com/PyCQA/isort/issues/2083


wj-Mcat (approved):
LGTM

PR #4825, opened 2023-02-16
[utc] set utc-base as default model

LemonNoel:
### PR types

Others

### PR changes

Others

### Description

- Set `utc-base` as default in Taskflow and modify corresponding docs.


ZHUI (approved):
LGTM

PR #4844, opened 2023-02-16
[AutoNLP] refactor verbosity for more effect log control

sijunhe:
### Description
- use `log_to_driver` in `ray.init` for more effective log control
- refactor verbosity to be at the constructor instead of at train
- add CsvLogger to avoid logging with Csv, json and tensorboardx by default


lugimzzz (approved):
lgtm

PR #4972, opened 2023-02-24
[Function optimization] add unittest for downloading with file-lock

wj-Mcat:
### PR types

Function optimization

### PR changes

APIs

### Description

添加文件下载时会 hang 住的单测，可是没有复现出问题。


sijunhe (approved):
lgtm

PR #5012, opened 2023-02-27
support feed inputs_embeds into the forward function of BertModel

FlyingQianMM:
### PR types

Function optimization

### PR changes

Models

### Description

We add the input variable `inputs_embeds` of the `forward` function of BertModel, in which users can feed the inputs_embeds manually.


sijunhe (commented):
thanks for the contribution!
Please enable the `use_test_inputs_embeds` flag in `bert/test_modeling.py` to confirm that inputs_embeds API works
https://github.com/PaddlePaddle/PaddleNLP/blob/develop/tests/transformers/test_modeling_common.py#L71

PR #5063, opened 2023-03-02
[SKEP P1] fix return params for SkepCrfForTokenClassification

1649759610:
### PR types

Function optimization

### PR changes

Models

### Description

fix return params for SkepCrfForTokenClassification


sijunhe (approved):
lgtm

PR #5128, opened 2023-03-07
torch_load use ZipFile, support bf16, support torch.nn.Parameter

JunnYu:
### PR types

Performance optimization
### PR changes

APIs
### Description

- 使用 ZipFile 读取torch权重
- 支持torch bfloat16
- 支持torch.nn.Parameter


sijunhe (approved):
lgtm

PR #5129, opened 2023-03-07
Fix mt5 config and add auto

gongel:
### PR types

Bug fixes
### PR changes

Models
### Description

- Fix mt5 config
- Support Auto load


sijunhe (approved):
ok

PR #5132, opened 2023-03-07
[DocParser] Add bbox check

linjieccc:
### PR types

Others
### PR changes

APIs
### Description

- Add bbox check that ensure the height and width of bbox are greater than zero


sijunhe (approved):
okay

PR #5183, opened 2023-03-10
copy newer_group form distutils.dep_util to avoid import distutils

guoshengCS:
### PR types

Others
### PR changes

Others
### Description

copy newer_group form distutils.dep_util to avoid import distutils


sijunhe (approved):
lgtm!

PR #5222, opened 2023-03-14
[OPT P1] add return_dict, labels and so on.

1649759610:
### PR types

 Function optimization 

### PR changes

Models

### Description

1. add return_dict, labels, output_attentions and output_hidden_states.
2. modify corresponding tests for opt.
3. update with config for OPTEmbedding and TransformerDecoder.


sijunhe (approved):
lgtm!

PR #5322, opened 2023-03-21
[GLM10B] Fix attention bug in MP mode

LemonNoel:
### PR types

Bug fixes

### PR changes

Models

### Description

- Fix variable name typos when cache is not None
- Fix attention shape error when calculating parallel_attention with tensor parallel model


sijunhe (approved):
lgtm

PR #5339, opened 2023-03-21
[GLM10B] Fix loss bug

LemonNoel:
### PR types

Bug fixes

### PR changes

Models

### Description

- Fix loss bug, which takes all non-pad token for loss calculation instead of using loss_mask


wawltor (approved):
LGTM

PR #5344, opened 2023-03-21
[LLM] low_cpu_mem_usage enable dtype guard

ZHUI:
### PR types

Function optimization
### PR changes

APIs
### Description

`low_cpu_mem_usage`  enable dtype guard


LemonNoel (approved):
LGTM

PR #5361, opened 2023-03-22
[GLM10B] Fix rouge calculation 

LemonNoel:
### PR types

Bug fixes

### PR changes

Others

### Description

- The prediction tokens were selected by `attention_mask` in previous version, and `padding` tokens were not ignored when calculate rouge scores, which leads to a lower metric number.


wawltor (approved):
LGTM

PR #5389, opened 2023-03-23
fix neg shape under to_static+prim mode

veyron95:
### PR types

Others
### PR changes

Others
### Description

fix neg shape under to_static+prim mode


JunnYu (approved):
LGTM

PR #5398, opened 2023-03-23
Fix sequence parallel bug in GPT model

sneaxiy:
### PR types
Bug fixes

### PR changes
Models

### Description
Fix sequence parallel bug in GPT model.


wawltor (approved):
LGTM

PR #5422, opened 2023-03-26
[Bug fixes] Update labelstudio2doccano.py

46319943:
### PR types
Bug fixes

### PR changes
Others. Helper function for convering the format of label-studio to doccano.

### Description

Fix json file reading error.
- Only json line format was accpted, which is not the default export option in labelstudio, making error for directly output json or formatted json file. The reading procedure was fixed to adapt both the json line format and detended format.

Add support for setting default value for missing relation type, which is allowed in label-studio but not the case in doccano.
- Some relation type could be emply list imported from label-studio. Add default relation type.

### DIFF
--- a/model_zoo/uie/labelstudio2doccano.py
+++ b/model_zoo/uie/labelstudio2doccano.py
@@ -109,7 +109,7 @@ if __name__ == "__main__":
     parser.add_argument(
         "--doccano_file",
         type=str,
-        default="doccano_ext.json",
+        default="doccano_ext.jsonl",
         help="Saving path in doccano format.",
     )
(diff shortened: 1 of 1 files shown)


shenhu365:
可以再提供一下doccano导出数据转换为Label Studio格式的代码吗？

46319943:
> 可以再提供一下doccano导出数据转换为Label Studio格式的代码吗？

可以在PaddleNLP下面提一个新的Issue

46319943:
@wawltor @sijunhe 请问能查看、通过一下吗？

PR #5428, opened 2023-03-27
[Typo] Update README.md

46319943:
### PR types
Typo

### PR changes
Change from single quotation marks to double quotation marks to avoid parsing error in Windows.


sijunhe (approved):
thanks

PR #5469, opened 2023-03-29
fix the parallel and recompute for bloom

wawltor:
### PR types
Bug fixes 

### PR changes
Models

### Description
fix the parallel and recompute for bloom


sijunhe (dismissed):
lgtm

PR #5541, opened 2023-04-06
Fix bugs when combining flash attention and core attention recompute

sneaxiy:
### PR types
Bug fixes

### PR changes
Others

### Description
Fix the bugs when combining flash attention and core attention recompute.


wawltor (approved):
LGTM

PR #5655, opened 2023-04-13
Remove dtype from __init__

LemonNoel:
### PR types

Bug fixes

### PR changes

APIs

### Description

- model_kwargs has an unexpected parameter `dtype` when `dtype` is set in `from_pretrained`.


sijunhe (commented):
see #5646

PR #5656, opened 2023-04-13
[LLM] Add predict generation and fix causal mask

linjieccc:
### PR types

Bug fixes
### PR changes

Models
### Description

- Fix causal mask


wawltor (approved):
LGTM

PR #5762, opened 2023-04-23
[LLM] Optimize llama: scale before matmul

linjieccc:
### PR types

Bug fixes
### PR changes

Models
### Description

- do query scale before matmul


wawltor (approved):
LGTM for the pure fp16 training

The BF16 is not need for the fix

PR #5866, opened 2023-05-08
fix the issue of ernie-layout model inference error cause by  invalid…

zirui:
### Description

fix the issue of ernie-layout model inference error cause by  invalid input of image (without ocr results), refer issue:#5865

### DIFF
--- a/model_zoo/ernie-layout/deploy/python/predictor.py
+++ b/model_zoo/ernie-layout/deploy/python/predictor.py
@@ -768,7 +768,7 @@ class Predictor(object):
                 invalid_idx.add(idx)
 
         # Save maping between original index in docs and new index(index in valid inputs)
-        idx_map = {j:i for i, j in enumerate(input_idx)}
+        idx_map = {j: i for i, j in enumerate(input_idx)}
 
         inputs = collections.defaultdict(list)
         for data in input_data:
@@ -790,7 +790,7 @@ class Predictor(object):
                 preds[1].extend(output[1].tolist())
         results = self.postprocess(preds)
 
-        # Merge result of invalid docs & valid docs input
+        # Merge results of invalid docs & valid docs input
         merged_results = []
         for i in range(len(docs)):
             if i in invalid_idx:
--- a/model_zoo/ernie-layout/deploy/python/predictor.py
+++ b/model_zoo/ernie-layout/deploy/python/predictor.py
@@ -754,12 +754,21 @@ class Predictor(object):
 
     def predict(self, docs):
         input_data = []
-        for doc in docs:
+        invalid_idx, input_idx, idx_map = set(), [], {}
+        for idx, doc in enumerate(docs):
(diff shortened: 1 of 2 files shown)


model_zoo/ernie-layout/deploy/python/predictor.py
```diff
@@ -754,12 +754,21 @@ def infer(self, data):
...
+        invalid_idx, input_idx, idx_map = set(), [], {}
+        for idx, doc in enumerate(docs):
             ocr_result = self.ocr.ocr(doc, cls=True)
             # Compatible with paddleocr>=2.6.0.2
             ocr_result = ocr_result[0] if len(ocr_result) == 1 else ocr_result
-            example = ppocr2example(ocr_result, doc)
-            input_data.append(example)
+            if ocr_result:
+                # Only process images with ocr results
+                example = ppocr2example(ocr_result, doc)
```
linjieccc: @zirui Thanks a lot for your contribution! 

I am thinking that if it would be better to handle the case of empty OCR results inside the ``ppocr2example`` function. 

The deployment script ``predictor.py`` is shared among NER, QA, and classification tasks. 

It would work fine for NER and QA tasks, but there might be issues if we handle it this way for the document classification task. If the OCR results are empty, they should still be assigned to a category. The current approach would result in an empty string.

zirui: > 
The NER/OA/classification tasks share the `predictor.py' script, so  all these tasks have this problem.
I agree that further discussion is needed on the return results when the OCR input is empty: What results should be returned for different tasks? Do you have any suggestions?

 "handle the case of empty OCR results inside the ppocr2example"  should also solve this problem, but it seems need more understanding about what  `example` to return  for subsequent processors of different  tasks(QA/NER/classification), so i choose the current simpler approach. I will further review the processing details of different tasks to confirm which method is more suitable

linjieccc: @zirui I would suggest handling the case of empty OCR results inside the ppocr2example function. 

https://github.com/PaddlePaddle/PaddleNLP/blob/develop/paddlenlp/utils/image_utils.py#LL733C5-L733C106

```python
example = {"text": doc_tokens, "bbox": doc_boxes, "width": im_w, "height": im_h, "image": img_base64}
```

For an example with empty OCR results, the "text" and "bbox" are expected to be empty lists, while the "width," "height," and "image" retain the original image information.

zirui: > 

I have adopted your suggestion to modify inside the `ppocr2example`, 
and now for empty ocr input,  the results returned by different tasks are similar to this:

1. cls: [{'doc': './images/test_image_no_ocr.png', 'result': 'specification'}]
2. mrc: []
3. ner: [{'doc': './images/test_image_no_ocr.png', 'result': []}]

linjieccc (approved):
LGTM. Thanks again!

PR #5872, opened 2023-05-08
[Bug fixes] disable tiny-random-bert checking

wj-Mcat:
### PR types

Bug fixes

### PR changes

Models

### Description

`tiny-random-bert` is different from hf-hub which will raise some error in unit testing, so disable md5value checking


sijunhe (dismissed):
ok

PR #5992, opened 2023-05-22
[LLM] return tensor for compatible with bf16

ZHUI:
### PR types

Function optimization
### PR changes

Others
### Description

return tensor for compatible with bf16


sijunhe (approved):
thanks

PR #6071, opened 2023-06-01
[arange] set dtype="int64" explicitly.

LemonNoel:
### PR types

Others

### PR changes

Others

### Description

- Set `dtype="int64"` in `paddle.arange` explicitly, as the behavior of `dtype=None` will change in next PaddlePaddle version.


ZHUI (approved):
LGTM

PR #6124, opened 2023-06-08
[AutoParallel] add sharding opt config

zhaoyinglia:
### PR types

Others
### PR changes

Others
### Description

- add sharding optimization config for 13B gpt model
- fix annotation for dp+pp


JZ-LIANG (approved):
LGTM

PR #6188, opened 2023-06-15
[model_zoo/gpt3] Fix import bug in finetune of GPT3

haohongxiang:
### PR types

Bug fixes
### PR changes

Others
### Description

[model_zoo/gpt3] Fix import bug in finetune of GPT3


ZHUI (approved):
LGTM

PR #6264, opened 2023-06-28
Disabled faster transformer& faster generation

zjjlivein:
### Description

Disabled faster transformer& faster generation 转至CE例行
add llama finetune case


sijunhe (approved):
lgtm

PR #6347, opened 2023-07-10
fix python include path

zhiqiu:
### PR types

Bug fixes
### PR changes

Others
### Description

In some case, `python -m pybind11 --includes` returns `/usr/local/include/python3.x/`, but the real header file is in `/usr/include/python3.x/`. So, add `python-config --includes`.
Ref: https://stackoverflow.com/questions/72502292/fatal-error-python-h-no-such-file-or-directory-when-compiling-pybind11-example


wawltor (approved):
LGTM

PR #6385, opened 2023-07-13
Refactor training loop merge meta save right away

liuzhenhai93:

### DIFF
--- a/paddlenlp/trainer/trainer.py
+++ b/paddlenlp/trainer/trainer.py
@@ -574,6 +574,30 @@ class Trainer:
                 raise ValueError(f"No valid checkpoint found in output directory ({self.args.output_dir})")
         return resume_from_checkpoint
 
+    def _load_check_point(self, resume_from_checkpoint, delay_optimizer_creation, max_steps):
+        # Check if saved optimizer or scheduler states exist
+        self._load_optimizer_and_scheduler(resume_from_checkpoint)
+        self.load_state_dict_from_checkpoint(resume_from_checkpoint)
+        model = self._wrap_model(self.model_wrapped)
+        # for the rest of this function `model` is the outside model, whether it was wrapped or not
+        if model is not self.model:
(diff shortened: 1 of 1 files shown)


paddlenlp/trainer/trainer.py
```diff
@@ -480,34 +480,41 @@ def recover_params_from_master_weights(self, state_dict):
...
-            paddle.distributed.all_gather_object(tmp, send_item, group=group)
-            for recv_item in tmp:
-                if recv_item[0] == "pack" and recv_item[1].size == 0:
-                    continue
-                if not filter_func(recv_item[0]):
-                    continue
-                gather_res[recv_item[0]] = paddle.to_tensor(recv_item[1], place=paddle.CPUPlace())
-                print(f"{i} gather {recv_item[0]} ")
-        return gather_res
+                tensor = paddle.to_tensor(np.zeros(shape, dtype))
```
sneaxiy, line 506: Can use `tensor = paddle.empty(shape, dtype=dtype)` to avoid init CPU zero numpy array.

paddlenlp/trainer/trainer.py
```diff
@@ -480,34 +480,41 @@ def recover_params_from_master_weights(self, state_dict):
...
+        state_dict = {k: map_func(v) for (k, v) in state_dict.items()}
+
+        meta_dict = {}
+        for (k, v) in state_dict.items():
+            # src rank
+            meta_dict[k] = (v.dtype, v.shape, group.rank)
+
+        meta_dict_list = self._all_gather_simple_object(meta_dict, group)
+        meta_list = [(k, v) for meta in meta_dict_list for (k, v) in meta.items()]
+        meta_list = sorted(meta_list, key=lambda x: x[1])
```
sneaxiy, line 499: Use `key=lambda x: x[0]` here.

Please check `key` is unique here. Add assertion here.

PR #6415, opened 2023-07-17
Support parsing json args

gongel:
### PR types

 New features
### PR changes

Others
### Description

- Support parsing json args,  like `python example.py params.json`


wawltor (approved):
LGTM

PR #6431, opened 2023-07-18
fix lr_scheduler use increment

zhaoyinglia:
### PR types

Bug fixes
### PR changes

Others
### Description

- fix lr_scheduler for auto parallel when use_increments is true


zhiqiu (approved):
LGTM

PR #6612, opened 2023-08-04
Fix ERNIE-1.0 static save

gongel:
### PR types

Bug fixes
### PR changes

Models
### Description

- Fix ernie static graph save


ZHUI (approved):
LTGM

PR #6649, opened 2023-08-08
[llama benchmark] Add to_static support for llama

w5688414:


sijunhe:
为啥做这个？llama这边的静态图我们不走这套，走FusedMT

PR #6721, opened 2023-08-14
[Llama benchmark]add llama benchmark scripts

wj-Mcat:
### PR types

New features

### PR changes

Models

### Description

添加 llama benchmark 测试脚本


wawltor (approved):
LGTM

PR #6856, opened 2023-08-29
[Function optimization] Update llama benchmark

wj-Mcat:
### PR types

Function optimization

### PR changes

Models

### Description

更新 llama benchmark


wawltor (approved):
LGTM

PR #6872, opened 2023-08-30
use pylayer version of recompute for llama

FeixLiu:
### PR types
Others

### PR changes
Others

### Description
Llama use pylayer version of recompute.

This pr: https://github.com/PaddlePaddle/Paddle/pull/56793 fix llama generation bug under pylayer version recompute.


sijunhe:
we will work on #6885 instead

PR #6904, opened 2023-09-04
Move intokens import  to the top of file

sijunhe:
### Description
because of https://github.com/PaddlePaddle/PaddleNLP/blob/develop/llm/finetune_generation.py#L313


lugimzzz (approved):
lgtm

PR #7187, opened 2023-10-10
[LLM] Fix error when config.bs is more than input bs.

xiaoxiaohehe001:
### PR types

Others
### PR changes

Others
### Description

 Fix error when config.bs is more than input bs.


wj-Mcat (approved):
LGTM

PR #7195, opened 2023-10-11
[Tokenizer] fix _decode

DrownFish19:
### PR types

Bug fixes
### PR changes

APIs
### Description

* bugs:
`sentencepiece` cannot process `numpy.ndarray`. 
Thus, when the data type of `token_ids`  is `numpy.ndarray`, tokenizer cannot decode it.

* solution:
`PretrainedTokenizer._decode`  requires `token_ids:List[int]` , thus `token_ids` will convert to list in `_decode`.

```python
class PretrainedTokenizer
    def _decode(
        self,
        token_ids: List[int],
        skip_special_tokens: bool = False,
        clean_up_tokenization_spaces: bool = True,
        spaces_between_special_tokens: bool = True,
        **kwargs
    ) -> str:
  
        if isinstance(token_ids, np.ndarray):
            token_ids = token_ids.tolist()
```
* fix issue #7172

### DIFF
--- a/paddlenlp/transformers/tokenizer_utils.py
+++ b/paddlenlp/transformers/tokenizer_utils.py
@@ -24,6 +24,7 @@ import unicodedata
 from collections import OrderedDict
 from typing import Dict, List, Optional, Tuple, Union
 
+import numpy as np
 import six
 from paddle.utils import try_import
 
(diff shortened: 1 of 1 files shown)


ZHUI (approved):
LGTM

PR #7269, opened 2023-10-18
fix error in dropout of hybrid_model

heavyrain-lzy:
### PR types

Bug fixes
### PR changes

Models
### Description

Fix the bug in `hybrid_model.py` in the case of 'not self.use_fused_dropout_add'.
Fixed the bug, test the hybrid_model using pp4-mp2. After 10 steps the losses are as follow:
|  scene   | loss  |
|  ----  | ----  |
| use_fused_dropout_add  | 11.255878448 |
| not use_fused_dropout_add(fixed bug)  | 11.255878448 |
| not use_fused_dropout_add(origin)  | 11.210844040 |


zhiqiu (approved):
LGTM

PR #7407, opened 2023-11-09
Fix dist dataloader with no mp group

DesmonDay:
### PR types
Bug fixes

### PR changes
Others

### Description
`mp_group is not None` cannot determine the existence of 'mp_group' correctly, use mp_group.nranks instead. The mp group build code:
``` 
if size > 1 and global_rank in ranks:
    rank = 0 if backend == 'heter' else ranks.index(global_rank)
    pg = _new_process_group_impl(
        backend,
        _default_store,
        rank,
        size,
        group_name,
        pg_options=None,
        group_id=gid,
    )
else:
    rank = -1
    pg = None
```

### DIFF
--- a/paddlenlp/data/dist_dataloader.py
+++ b/paddlenlp/data/dist_dataloader.py
@@ -159,7 +159,7 @@ class DistDataLoader(paddle.io.DataLoader):
 
         # Broadcast data keys size.
         if self._data_keys_size is None:
-            if self.mp_group is not None and self.pp_rank == 0:
+            if self.mp_group.nranks > 1 and self.pp_rank == 0:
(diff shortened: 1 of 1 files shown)


paddlenlp/data/dist_dataloader.py
```diff
@@ -159,7 +159,7 @@ def __next__(self):
 
         # Broadcast data keys size.
         if self._data_keys_size is None:
-            if self.mp_group is not None and self.pp_rank == 0:
+            if self.mp_group.rank > -1 and self.pp_rank == 0:
```
ZHUI:
```suggestion
            if self.mp_group.nranks > 1 and self.pp_rank == 0:
```

DesmonDay: 其实都可，已改

ZHUI (approved):
LGTM

PR #7455, opened 2023-11-16
Add TP delay scale loss and TP sharding comm overlap

sneaxiy:
### PR types
Function optimization

### PR changes
Others

### Description
Add TP delay scale loss and TP sharding comm overlap.


ZHUI (approved):
LGTM

PR #7509, opened 2023-11-23
[Trainer] Remove dist_gather, use `gather` under paddle.distributed instead

DesmonDay:
### PR types
Bug fixes

### PR changes
Others ] -->

### Description
Original `dist_gather` function may fail in some cases, use paddle.distributed.communication.stream.gather instead.


ZHUI (approved):
LGTM

PR #7582, opened 2023-12-04
[Dist Dataloader] Fix dist dataloader sampler

DesmonDay:
### PR types
Bug fixes

### PR changes
Others

### Description
To ensure when using distributed dataloader, no matter the process has dataset or not, the type of train_sampler is the same.


wawltor (approved):
LGTM

PR #7617, opened 2023-12-08
[Pretrain] Fix Qwen paddle.outer calculation

DesmonDay:
### PR types
Performance optimization

### PR changes
Others

### Description
Since `paddle.outer` is composed of `matmul`, and `matmul` is added to amp_white_list. When the two inputs are `float32` dtype, the calculation is still under amp, which may cause loss of accuracy.


ZHUI (approved):

wawltor (approved):
LGTM

PR #7619, opened 2023-12-08
Type improvement behavior correction, index_sample API already support float16, logits no need to cast. 

zxcd:
### PR types

Bug fixes
### PR changes

Others
### Description

Type improvement behavior correction, index_sample API already support float16, logits no need to cast.


wj-Mcat (approved):
LGTM

PR #7658, opened 2023-12-14
【AutoParallel】Support 'master_grad' in Llama in static auto-parallelism

heavyrain-lzy:
### PR types

Others
### PR changes

Others
### Description

Add `master_grad` optional in static auto-parallelism.You can set `--amp_master_grad 1` to enable master_grad in AMP-O2.


wawltor (approved):
LGTM

PR #7746, opened 2023-12-29
[AutoParallel] Add Sequence Parallel for Static LLaMA

JZ-LIANG:
### PR types
Function optimization

### PR changes
Others

### Description
Add Sequence Parallel for Static LLaMA


wawltor (approved):
LGTM

PR #7772, opened 2024-01-04
Fix shared weights sync for PipelineLayer 

DrownFish19:
### PR types

Bug fixes
### PR changes

Others
### Description

1. PipelineLayer will create parameters for each layer and call `_synchronize_shared_weights()` to synchronize the shared parameters.
2. When setting the model `state_dict`, `_synchronize_shared_weights` will be called to synchronize the shared parameters. However, `self._init_weights` will re-initialize the parameters without synchronizing the shared parameters. If the following step does not load a checkpoint, the shared parameters will be different.


ZHUI (approved):

wawltor (approved):
LGTM

PR #7794, opened 2024-01-08
[unified checkpoint] fix checkpoint names

DrownFish19:
### PR types

Bug fixes
### PR changes

Others
### Description

1. save master weights name as model weights name when SKIP_SAVE_MODEL_WEIGHT.
2. fix config save skip when SKIP_SAVE_MODEL_WEIGHT.


ZHUI (approved):
LGTM

PR #7806, opened 2024-01-09
[Pretrain] Fix eval during pretrain

DesmonDay:
### PR types
Bug fixes

### PR changes
Others

### Description
There might be some problems when using numpy share memory under multi-processing situation. Therefore we directly copy the data to paddle.Tensor in `_collate_data` func.


ZHUI (approved):
LGTM

PR #7815, opened 2024-01-10
[LLM] fix Qwen-7b-Chat precision issue

ziangqin-baidu:
set default tokenizer padding style as "left" in one of existing code branch

### PR types

Bug fixes

### PR changes

llm/predictor.py > create_predictor()

### Description

Fix Qwen-7B-Chat model precision issue, in which batch inference return false results.
The issue is caused by error setting of tokenizer `padding_side`, which should be "left" instead of default "right".
This PR modifies default `padding_side` as "left" in one of tokenizer creation code branch to fix the issue.


wj-Mcat (approved):
LGTM

PR #7844, opened 2024-01-12
[text_classification_retrieval_based] fix bug of evaluate.py

ZeyuTeng96:
Hi there,

fix same bug as evaluate.py in neural search.

Thanks


w5688414 (approved):
LGTM

PR #7874, opened 2024-01-21
【AutoParallel】Support bf16 loss in static

heavyrain-lzy:
### PR types

Others
### PR changes

Others
### Description

Support to log loss with dtype `bfloat16` aligning to dygraph https://github.com/PaddlePaddle/PaddleNLP/blob/2a8d1380fa98fa5e5dd2eee8943cfe42660198c8/paddlenlp/trainer/trainer.py#L2649


zhiqiu (approved):
LGTM

PR #7882, opened 2024-01-22
[Bug Fix] fix slice bug in LlamaRotaryEmbedding

MarioLulab:
### PR types
Bug fixes

### PR changes
Others

### Description
In `LlamaRotaryEmbedding`, `self.cos_cached` and `self.sin_cached` are both the shape of `[1, max_seq_len_cached, 1, self.dim]`, but `forward` function wrongly does slice operation on dimension 2 which is the dimension with the constant value `1`. We update dimensions which should be slice actually.

Snipaste_2024-01-23_16-30-55


MarioLulab:
可以帮忙 review 一下吗 😃  @wawltor @ZHUI

lugimzzz (approved):
lgtm

wawltor (approved):
LGTM

PR #7886, opened 2024-01-23
Change Paddle.fluid to Paddle.base

ATblackwhite:
### PR types

Version Update
### PR changes

Models
### Description

Since the upgrade of Paddle, Paddle.fluid has been discontinued and replaced with Paddle.base. So I changed all the places where fluid was used in the file to base

### DIFF
--- a/paddlenlp/trainer/trainer.py
+++ b/paddlenlp/trainer/trainer.py
@@ -176,7 +176,7 @@ except:
 try:
     from paddle.io.dataloader.dataloader_iter import _DataLoaderIterBase
 except:
-    from paddle.fluid.dataloader.dataloader_iter import _DataLoaderIterBase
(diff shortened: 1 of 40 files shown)


paddlenlp/trainer/trainer.py
```diff
@@ -176,7 +176,7 @@
 try:
     from paddle.io.dataloader.dataloader_iter import _DataLoaderIterBase
 except:
-    from paddle.fluid.dataloader.dataloader_iter import _DataLoaderIterBase
+    from paddle.base.dataloader.dataloader_iter import _DataLoaderIterBase
```
ZHUI, line 179: api 的位置都对过吗？

ZHUI (changes requested):

PR #7895, opened 2024-01-25
[LLM] fix llama precision on custom devices

SylarTiaNII:
### PR types

Bug fixes
### PR changes

Models
### Description

Remove redundant type casting on masked_lm_loss which may cause precision failures on custom devices


ZHUI (approved):
LGTM

Issue #7902, opened 2024-01-25
[Bug]:  ernie tiny deploy python infer_model.py  get wrong slot with chinese、number or english chars in query


maowandong:
### 软件环境

```Markdown
- paddlepaddle:
- paddlepaddle-gpu: 2.5.2.post120
- paddlenlp: 2.6.1
```

### 重复问题

- [X] I have searched the existing issues

### 错误描述

```Markdown
query: "拨打电话18866668888"

preds: [46, 46, 47, 47, 47, 12, 13, 13, 13, 13]
No. 3 text = 打电话给18866668888
{'intent': 'App.Phone', 'confidence': 0.9999325, 'slot': [{'slot': 'action', 'entity': '打电话给', 'pos': [0, 3]}]}

preds:    [46, 46, 47, 47, 47, 12, 13, 13, 13, 13]
labeling: [B-action, B-action, I-action, I-action, I-action, B-number, I-number, I-number, I-number, I-number]

the preds is align with tokenizer results, not the query. So，if the query is mixed with chinese chars and numbers or english chars, it will not segmented by char.
```

### 稳定复现步骤 & 代码

query: "拨打电话18866668888"

w5688414:
thank you for your contribution

PR #7910, opened 2024-01-26
[Bug]: ernie tiny deploy python infer_model.py get wrong slot with ch…

maowandong:
…inese、number or english chars in query #7902

### PR types

Bugfix

[Bug]: ernie tiny deploy python infer_model.py get wrong slot with chinese、number or english chars in query #7902

### PR changes

python infer_model.py cls_postprocess

original: align preds out with queries;
fixed: align preds with tokenizer's token, then labeling tokens  and tokenizer decode to get orignal chars.


w5688414:
Thank you, Please fix the conflicts first!

PR #7965, opened 2024-02-05
[npu] llama modeling support fa

tianhaodongbd:
### PR types

Others
### PR changes

Others
### Description

Others

### DIFF
--- a/paddlenlp/transformers/llama/modeling.py
+++ b/paddlenlp/transformers/llama/modeling.py
@@ -16,6 +16,7 @@
 from __future__ import annotations
 
 import math
+import os
 import warnings
 from functools import partial
 from typing import Optional, Tuple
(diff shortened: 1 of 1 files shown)


paddlenlp/transformers/llama/modeling.py
```diff
@@ -189,7 +195,6 @@ def scaled_dot_product_attention(
 ):
     bsz, q_len, num_heads, head_dim = query_states.shape
     _, kv_seq_len, _, _ = value_states.shape
```
gongweibao, line 197:
```
if config.use_flash_attenion and not flash_attension:
    assert False, "flash_attension OP is not ready.Please check your env"
```

gongweibao (changes requested):
Don't modify the original GPU logic.

PR #8048, opened 2024-03-04
Fit pir flag in predictor

cyber-pioneer:
### PR types
Bug fixes

### PR changes
Others

### Description
To make pir executor mode come into effect.


wj-Mcat (approved):
LGTM

PR #8049, opened 2024-03-04
fix the index_dataset with old data format

wawltor:
### PR types
Bug fixes 

### PR changes
 Models

### Description
fix the index_dataset with old data format


ZHUI (approved):
LGTM

PR #8199, opened 2024-03-27
【AutoParallel】Change the `dtype` of initializing the model

heavyrain-lzy:
### PR types

Others
### PR changes

Others
### Description

The `amp` of static graph model can't accept a model initialized with `dtype float16 or bfloat16`, so use the `dtype float32` to initialize the model in static graph model.


zhiqiu (approved):
LGTM

PR #8259, opened 2024-04-11
fix autoruner resume case

Difers:
### PR types

Others
### PR changes

Others
### Description

fix autoruner resume case


ZHUI (approved):
LGTM

PR #8360, opened 2024-05-06
fix bug for load pretrained sharded model when train with lora

myjlyjly:
### PR types

Bug fixes

### PR changes

Others

### Description

When train the lora model and load from full-parameter pretrained sharded model, we need to delete the fixed parameter in optimizer.

### DIFF
--- a/paddlenlp/trainer/trainer.py
+++ b/paddlenlp/trainer/trainer.py
@@ -383,7 +383,7 @@ class Trainer:
         if self.args.pipeline_parallel_degree > 1:
             from paddle.distributed.fleet.meta_parallel import PipelineLayer
 
-            assert isinstance(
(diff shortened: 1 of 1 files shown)


paddlenlp/trainer/trainer.py
```diff
@@ -521,6 +521,22 @@ def load_state_dict_from_checkpoint(self, resume_from_checkpoint=None):
             state_dict = self.load_state_dict_from_checkpoint_with_reshard(resume_from_checkpoint)
             if self.args.bf16:
                 state_dict = self.recover_params_from_master_weights(state_dict)
+
+            for p in self.model.parameters():
```
ZHUI, line 525: 这里有判断哪些参数是冻住的吗?

myjlyjly: 当时只是调通 lora 了, 确实没注意, 我再完善一下...

PR #8378, opened 2024-05-07
[LLM] Fix fuse or split with same key

DrownFish19:
### PR types

Bug fixes

### PR changes

Others

### Description

1.Fix fuse, split or other convert actions with same key.

**bug**: when the origin and converted keys are the same, paddlenlp will split the origin state_dict and the converted state_dict for tensor parallel twice. 

**fix**: if the origin and converted keys are the same, paddlenlp should only split the origin state_dict for tensor parallel.

2. change the diff eps to 1e-5


ZHUI (approved):
LGTM

PR #8535, opened 2024-06-03
fix sharding reshard save

sneaxiy:
### PR types
Bug fixes

### PR changes
Others

### Description
Fix sharding reshard save when the model is not `PretrainedModel`.


ForFishes (approved):
LGTM

PR #8543, opened 2024-06-05
Save parameter shape and dtype when using sharding reshard

sneaxiy:
### PR types
Function optimization

### PR changes
Others

### Description
Save parameter shape and dtype when using sharding reshard.


ForFishes (approved):
LGTM

ForFishes (approved):
LGTM

PR #8546, opened 2024-06-05
Fix memory leak bug

sneaxiy:
### PR types
Bug fixes

### PR changes
Others

### Description
Fix memory leak bug of PaddleNLP.


ZHUI (approved):
LGTM

PR #8587, opened 2024-06-12
fix autotunner benchmark error and fix llama2 dy2st benchmark

fightfat:
### PR types
Bug fixes 

### PR changes
Others 

### Description
fix autotunner benchmark error and fix llama2 dy2st benchmark


ZHUI (approved):
LGTM

PR #8667, opened 2024-06-26
【AutoParallel】Change benchmark config for llama2-7b

heavyrain-lzy:
### PR types

Others
### PR changes

Others
### Description

Enable to use `reduce_agv` to replace `reduce_sum + scale`.


zhiqiu (approved):
LGTM

PR #8750, opened 2024-07-11
[AutoParallel] update loss for global clip

JZ-LIANG:
### PR types

Performance optimization
### PR changes

Others
### Description

update loss for global clip


zhiqiu (approved):
lgtm

PR #8768, opened 2024-07-16
Add the missing truncation=True in llm/predictor.py

lszxb:
### PR types

Bug fixes

### PR changes

Others

### Description

In `llm/predictor.py`, two calls to `tokenizer` are missing the `truncation=True` parameter, causing the results not to be truncated to `max_length`. This leads to errors in subsequent processing when using `--benchmark` for testing. This PR fixes this issue.


vivienfanghuagood (approved):
Approved

DesmonDay (approved):
LGTM

PR #8769, opened 2024-07-16
fix incorrect token counting in `llm/predictor.py`

lszxb:
### PR types

Bug fixes

### PR changes

Others 

### Description

The `benchmark()` function is currently counting the number of generated characters instead of the generated tokens, which is incorrect for speed test. This PR modified the `predict()` interface and let it can optionally return the generated tokens before postprocessing, so that it can count the number of generated tokens in the `benchmark()` function.

### DIFF
--- a/llm/predict/predictor.py
+++ b/llm/predict/predictor.py
@@ -248,16 +248,19 @@ class BasePredictor:
     def _infer(self, inputs):
         raise NotImplementedError
 
-    def _postprocess(self, predictions):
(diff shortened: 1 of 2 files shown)


vivienfanghuagood:
@RichardWooSJTU  @yuanlehome

vivienfanghuagood (approved):

DesmonDay (approved):
LGTM

PR #8791, opened 2024-07-22
fix version show

DrownFish19:
### PR types

Others
### PR changes

Others
### Description

#8754 makes the style of the version `3.0.0b0+a3223317` (`version+commit id`). However, that does not allow us to get the latest version from `https://www.paddlepaddle.org.cn/whl/paddlenlp.html` based on the latest date. Thus, we have modified the version style to `3.0.0b0.dev20240722` (`version+datetime`).

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -15,6 +15,7 @@ import errno
 import io
 import os
 import subprocess
+from datetime import datetime
 
 import setuptools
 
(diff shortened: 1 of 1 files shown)


ZHUI (approved):
LGTM

PR #8792, opened 2024-07-23
update version

DrownFish19:
### PR types

Others
### PR changes

Others
### Description

#8754 and #8791 make the style of the version  `3.0.0b0.dev20240723` (version+datetime). However, that does not allow us to get the latest version from https://www.paddlepaddle.org.cn/whl/paddlenlp.html based on the latest date. Thus, we have modified the version style to `3.0.0b0.post202407223` (version+post+datetime).

### DIFF
--- a/paddlenlp/__init__.py
+++ b/paddlenlp/__init__.py
@@ -14,6 +14,7 @@
 
 import os
 import sys
+from datetime import datetime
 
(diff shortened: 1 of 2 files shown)


wawltor (approved):
LGTM

PR #8803, opened 2024-07-24
[Trainer] Add enable_sp_async_reduce_scatter

DesmonDay:
### PR types
Others

### PR changes
Others

### Description
Add `enable_sp_async_reduce_scatter` for sequence parallel, only works when using tensor parallel and sequence parallel.


ZHUI (approved):
LGTM

PR #8811, opened 2024-07-25
support Llama3.1 8B 128K generation on single GPU 80GB

GuoxiaWang:
### PR types

Function optimization
### PR changes

Models
### Description

support Llama3.1 8B 128K generation on single GPU 80GB


wawltor (approved):
LGTM

PR #8851, opened 2024-07-31
Make sharding reshard compatible with wrong master weight names

sneaxiy:
### PR types
Bug fixes

### PR changes
Others

### Description
Make sharding reshard compatible with wrong master weight names.


ForFishes (approved):
LGTM

PR #8918, opened 2024-08-12
fix pipeline eval

gongel:
### PR types

Bug fixes
### PR changes

Others
### Description

- When do_train=False, do_eval=True, we need to wrap model for pipeline


ZHUI (approved):
LGTM

PR #8943, opened 2024-08-15
[Distributed]Add loss nan/inf checker

ForFishes:
### PR types

Breaking changes
### PR changes

Others
### Description

[Distributed]Add loss nan/inf checker

### DIFF
--- a/paddlenlp/trainer/trainer.py
+++ b/paddlenlp/trainer/trainer.py
@@ -995,7 +995,7 @@ class Trainer:
                 else:
                     tr_loss_step = self.training_step(model, inputs)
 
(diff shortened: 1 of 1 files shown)


paddlenlp/trainer/trainer.py
```diff
@@ -995,6 +995,10 @@ def _inner_training_loop(
                 else:
                     tr_loss_step = self.training_step(model, inputs)
 
+                if self.amp_dtype != "float16":
```
ZHUI: ci 报错了，需要先判断， ` if args.fp16 `

ForFishes: ok, i will fix it

PR #8974, opened 2024-08-21
[Unified checkpoint] update optimizer async save signal

DesmonDay:
### PR types
Bug fixes

### PR changes
Others

### Description
Update optimizer async save signal.


wawltor (approved):
LGTM

PR #9019, opened 2024-08-27
[Performance] Compatible with flashmask API rename upgrade

GuoxiaWang:
### PR types

 Function optimization
### PR changes

APIs
### Description

compatible with flashmask API rename upgrade


wawltor (approved):
LGTM

PR #9056, opened 2024-08-30
【Fix】Initialize dp degree in single GPU

greycooker:
### PR types
Fix

### PR changes
Initialize dp degree in single GPU

### Description
Initialize dp degree in single GPU


ZHUI (approved):
LGTM

PR #9102, opened 2024-09-06
[CI] Fix paddlepaddle install

DesmonDay:
### PR types
Bug fixes

### PR changes
Others

### Description
Fix paddlepaddle install.

PR #9149, opened 2024-09-18
fix llm ce predict run error

fightfat:
### PR types
Bug fixes

### PR changes
Others

### Description
fix llm ce predict run error because of #9094 and #9102


ZHUI (approved):
LGTM

PR #9274, opened 2024-10-15
[Unified Checkpoint] update async save logic

DesmonDay:
### PR types
Others

### PR changes
Others

### Description
1. Add new training argument, `output_signal_dir`, which is used to save asynchronous saving signal.
2. Fix async saving hang, when the asynchronous process is exited, raise RuntimeError directly.


wawltor (approved):
LGTM

PR #9290, opened 2024-10-18
[Adam/AdamW] Update adamw.py

HydrogenSulfate:
### PR types

Function optimization

### PR changes

Others 

### Description

`amsgrad` will be supported in <https://github.com/PaddlePaddle/Paddle/pull/68079> and added into parameter list of those two optimizers, therefore number of return value of `_C_ops.XX` will be incresed by 1. So `_` need to be replaced with `*_` for `compatibility`.


HydrogenSulfate:
closed as #9568 has been merged

PR #9310, opened 2024-10-24
fix non-pipelinelayer to distributed

gongel:
### PR types

Bug fixes
### PR changes

APIs
### Description

- When model is not pipelinelayer, it will assert error.

### DIFF
--- a/paddlenlp/trainer/trainer.py
+++ b/paddlenlp/trainer/trainer.py
@@ -2927,8 +2927,10 @@ class Trainer:
         prediction_loss_only = prediction_loss_only if prediction_loss_only is not None else args.prediction_loss_only
 
         if self.args.pipeline_parallel_degree > 1:
+            from paddle.distributed.fleet.meta_parallel import PipelineLayer
+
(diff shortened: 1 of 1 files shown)


paddlenlp/trainer/trainer.py
```diff
@@ -2927,8 +2927,10 @@ def evaluation_loop(
         prediction_loss_only = prediction_loss_only if prediction_loss_only is not None else args.prediction_loss_only
 
         if self.args.pipeline_parallel_degree > 1:
```
ZHUI, line 2929: 前面 有个 if self.args.pipeline_parallel_degree > 1: 的判断，非PP的模型也能进去？

ZHUI (approved):

ZHUI:
PPO情况下，主模型是 pp, dummpy的模型非 pp 导致的问题

paddlenlp/trainer/trainer.py
gongel: PPO有个dummy的model也会进去

PR #9340, opened 2024-10-30
align default custom black/white list for dygraph and static graph

zhiqiu:
### PR types

Bug fixes
### PR changes

Models
### Description

align default custom black/white list for dygraph and static graph


wawltor (approved):
LGTM

PR #9347, opened 2024-10-31
Add ordered save to avoid OOM

ForFishes:
### PR types

New features
### PR changes

Others
### Description

Add ordered save to avoid OOM


wawltor (approved):
LGTM

PR #9449, opened 2024-11-18
Fix QuantLora param init bug

will-jl944:
### PR types
Bug fixes

### PR changes
Models

### Description
Fix quant LoRA layers parameter initialization bug.


wawltor (approved):
LGTM

PR #9452, opened 2024-11-19
fixed the init problem in tensor parallel

wawltor:
### PR types
 Bug fixes

### PR changes
Models
### Description
fixed the init problem in tensor parallel


DrownFish19 (approved):
LGTM

PR #9456, opened 2024-11-19
【Auto-Parallel】update conf for sharding overlap in static

liym27:
### PR types

Performance optimization 

### PR changes

Models

### Description

update conf for sharding overlap in static


zhangbo9674 (approved):
LGTM

PR #9467, opened 2024-11-20
support offload/reload optimizer's states for custom device

tianhaodongbd:
### PR types

Others
### PR changes

Others
### Description

support offload/reload optimizer's states for custom device


gongweibao (approved):
LGTM

PR #9495, opened 2024-11-26
[XPU] set appropriate mask value for xpu

runzhech:
### PR types
Bug fixes

### PR changes
Others

### Description
Set appropriate mask value for xpu


wawltor (approved):
LGTM

PR #9534, opened 2024-11-29
[CI] Compatible with paddle.where

DrownFish19:
### PR types

Function optimization
### PR changes

APIs
### Description

Compatible with `paddle.where`, which only accepts tensors of bool dtype.


FlamingoPg:
QAQ blocking here

blacksheep-Aristotle:
> QAQ blocking here

+1

PR #9541, opened 2024-12-02
add llama3.1 config for auto_parallel

zhiqiu:
### PR types

Others
### PR changes

Models
### Description

add llama3.1 8b config for auto_parallel
The trainning speed on H GPU is 6984 tpsc on average
image


From00 (approved):
LGTM

PR #9549, opened 2024-12-03
[XPU] qwen2 supports flash_attn on XPU

will-jl944:
### PR types
New features

### PR changes
Models

### Description
qwen2 supports flash_attn on XPU


wawltor (approved):
LGTM

PR #9616, opened 2024-12-11
Optimize performance for Qwen2 model

sneaxiy:
### PR types
Performance optimization

### PR changes
Others

### Description
Optimize the performance of Qwen2 model.


wawltor (approved):
LGTM

PR #9629, opened 2024-12-13
Update embedding trainer state

DesmonDay:
### PR types
Function optimization

### PR changes
Others

### Description
Update embedding trainer state, add consumed_samples.


wawltor (approved):
LGTM

PR #9656, opened 2024-12-18
[Trainer] Fix loading rng state

DesmonDay:
### PR types
Bug fixes

### PR changes
Others

### Description
Hybrid paralell rng states changed when training environment differs, so we dot not set state tracker here.


ZHUI (approved):
LGTM

PR #9685, opened 2024-12-24
Set tensor parallel name mapping when fusion is used

sneaxiy:
### PR types
Function optimization

### PR changes
Others

### Description
Set tensor parallel name mapping when fusion is used.


wawltor (approved):
LGTM

PR #9707, opened 2024-12-27
[Inference] Use cuda core(int8_sq) for m <=4 in gemm_dequant OP

zhink:
### PR types

Performance optimization
### PR changes

APIs
### Description

use cuda core(int8_sq) for m <=4 in gemm_dequant OP
speedup 4.7% in decoder with batchsize=1


ZHUI (approved):
LGTM

PR #9757, opened 2025-01-08
[Trainer] update sequence parallel

DesmonDay:
### PR types
Function optimization

### PR changes
Others

### Description
When use pipeline parallel and sequence parallel simultaneously, we should turn off partial send recv.


ZHUI (dismissed):
LGTM

PR #9779, opened 2025-01-14
fix pickle unsafe-load

DrownFish19:
#### Before submitting

```shell

pre-commit run --file XXXX.py
```

### PR types

Function optimization
### PR changes

Others
### Description

Fix the unsafe loading issue with pickle by defining a SafeUnpickler that only allows loading data types in {"int", "float", "str", "tuple", "list", "dict", "set"}.


wawltor (approved):
LGTM

PR #9929, opened 2025-02-24
[‌CI]Update CI scripts

Liujie0926:
#### Before submitting

```shell

pre-commit run --file XXXX.py
```


csrc/setup.py
```diff
@@ -37,6 +37,13 @@ def read_version():
     """
     read version and return content
     """
+    VERSION = "3.0.0b3.post"
```
DrownFish19: 这个要不要考虑直接拿paddlenlp的版本？

PR #10021, opened 2025-03-07
[Performance optimization] Sage Attention Enhancement

l1cacheDell:
### PR types

Performance optimization

### PR changes

APIs

### Description

This PR is aiming at enhancing the usage of Sage Attention kernels. Including the following changes:

+ Fuse sage attention kernels
+ Forming a unified entry, avoid python-based organized form of invoking kernels.
    + remove `sageattention.py`
+ Remove adunbdant kernels & operations. Something like `lse` tensor computing was removed from kernel (we won't use that temperarily), which may improve the performance.


l1cacheDell:
Tested on both dynamic & static mode. Correct answer ensured.

chang-wenbin (approved):

yuanlehome (approved):

PR #10360, opened 2025-04-08
[Feature] Sage Attention Support Varlen & write kv cache BF16/FP16 

l1cacheDell:
### PR types
New features

### PR changes

APIs

#### 1. write kv cache (BF16/FP16) with RoPE integration
This PR add new feature with Sage Attention and integrating with write kv cache kernels, offering flexible cpp operator API like append attention, easy to use.

#### 2. Varlen support, allow different seqlens in a single batch!
This PR also add varlen support for different sequence length in a single batch processing. The key solution is to enlarge the `gridDim` (The amount of blocks when launching) to `[max_seqlen, num_heads, bsz]`, where `max_seqlen` is the longest sequence length in this batch. 

In kernel implementation, we allow the rest threads in the same blocks (processing the edge of a sequence) executing to avoid accuracy loss (which will introduce a bit extra latency, but it's necessary). Return and end up the kernel processing when the `thread_base_token` exceeds the current sequence length. 

Due to the Sage Attention kernel implementation, we need to support **varlen** feature not only for attn forward kernels, but also for quantization kernels (quant qk, quant_transpose_permute v kernels), *which takes extra developing & debugging time more than estimated.*


l1cacheDell:
image

Varlen supported! Open for review!

Tested & Verified Model:

+ Qwen2-1.5B-Instruct
+ Qwen2.5-14B-Instruct

l1cacheDell:
image

SM89 supported!

l1cacheDell:
image

Kernel validated varlen feature on SM80 arch devices.

PR #10537, opened 2025-05-06
[LLM] enhance logger output for real filename and lineno

SylarTiaNII:
### PR types

Function optimization

### PR changes

Others

### Description

[Pcard-90403]
CP from: https://github.com/PaddlePaddle/PaddleNLP/pull/9630
enhance logger output for real filename and lineno


wawltor (approved):
LGTM

PR #10682, opened 2025-05-30
Revert "[Auto-Parallel] optimize llama27b to avoid unnecessary communication"

liym27:
Reverts PaddlePaddle/PaddleNLP#10671


wawltor (approved):
LGTM

PR #10842, opened 2025-07-11
[RL] fix qwen load when fuse is enabled and modify gate precision to fp32

SylarTiaNII:
### PR types

Bug fixes

### PR changes

Models

### Description

[Pcard-90403]
Some fix for qwen2-moe & qwen3-moe
- fix load when fuse is enabled (fuse_qkv & fuse_ffn)
- modify gate precision to fp32


aiyinyuedejustin (approved):
OK

sneaxiy (approved):

PR #10875, opened 2025-07-21
[LLM] fix moe using on tensor parallelism

SylarTiaNII:
### PR types

Bug fixes

### PR changes

Others

### Description

[Pcard-90403]
fix moe using on tensor parallelism


DrownFish19 (approved):
LGTM

PR #11048, opened 2025-09-02
[fix] Broadcast optimizer state using broadcast_dp without shard-resh…

bo-ke:
…ard.

#### Before submitting

```shell

pre-commit run --file XXXX.py
```


bo-ke:
sync

PR #11051, opened 2025-09-02
Cherry-pick useful PRs from fleety

SylarTiaNII:
### PR types

Others

### PR changes

Others

### Description

- ZCC supports saving rng states
- Optimize ZCC save time
- Qwen-A3B supports TP-MoE
- Qwen-A3B supports fused ffn and attn
- Increase gate precision of Qwen-A3B to FP32
- fix normal save load
- support disable paddle api monkey patch by FLAG
- use safer grad sync when enabling sequence parallel


From00 (approved):
LGTM

PR #11073, opened 2025-09-05
Support to load sharded EMA checkpoint

sneaxiy:
#### Before submitting

```shell

pre-commit run --file XXXX.py
```

### PR types
Function optimization

### PR changes
Others

### Description
Support to load sharded EMA checkpoint.

**Note: Users should notice that the non-EMA checkpoint cannot be deleted when loading EMA checkpoint.**


From00 (approved):
LGTM

PR #11103, opened 2025-09-22
[Distributed] refine training_pipeline_step to avoid memory leaks

zhangyuqin1998:
### PR types

Function optimization

### PR changes

Others

### Description

Pass a local function to forward_backward_pipeline instead of the dataset itself. This prevents the dataset from being passed as a direct argument to forward_backward_pipeline, which would create additional reference counts that cannot be cleared, leading to GPU memory leaks.

This PR depends on https://github.com/PaddlePaddle/Paddle/pull/75446.


ForFishes (approved):
LGTM

sneaxiy (approved):

PR #11152, opened 2025-10-27
LoRA: add lora target modules support for qwen2 and qwen3 dense model…

aiyinyuedejustin:
LoRA: add lora target modules support for qwen2 and qwen3 dense model when using fused keys


lugimzzz (approved):
LGTM

PR #11153, opened 2025-10-27
LoRA: add lora target modules support for qwen2 and qwen3 dense model…

aiyinyuedejustin:
LoRA: add lora target modules support for qwen2 and qwen3 dense model…


lugimzzz (approved):
LGTM

From00 (approved):

PR #11246, opened 2026-08-02
docs(inference): note OpenAI-compatible multi-model gateway example (DaoXE)

seven7763:
## Summary

The English inference quick-start already shows calling the local Flask server with the OpenAI Python client. This PR adds a short note that the same client pattern works against any OpenAI-compatible multi-model gateway — DaoXE (`https://api.daoxe.com/v1`) is one concrete example.

Docs only.

## Test plan
- [ ] Local OpenAI client example unchanged
- [ ] Gateway note is clearly optional

Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>

### DIFF
--- a/docs/en/get_started/inference.md
+++ b/docs/en/get_started/inference.md
@@ -73,6 +73,16 @@ if stream:
 else:
     print(completion.choices[0].message.content)
 ```
+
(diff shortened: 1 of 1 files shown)


seven7763:
This is an outside docs addition that would need a CLA signature to go anywhere, and it isn't worth that process on either side. Closing it. Sorry for the queue time.

